PR #102, opened 2026-03-21
fix: terminal view send blocked + global shortcuts fire in /#terminal

meawpipatgmail:
## Summary

- **TerminalView send silently blocked**: `handlers.ts` guards sends with `/claude|codex|node/i` regex, but Claude Code reports its process name as a version number (e.g. `2.1.81`) — not `claude`. Added `force: true` to `TerminalView`'s send message to bypass the guard, since `/#terminal` is a raw terminal view that should work with any pane.
- **Global shortcuts fire while typing in terminal**: `j` → jump overlay, `v` → vs view, `i` → inbox, `/` → jump all triggered while typing in the terminal input div. The global handler only excluded `HTMLInputElement`/`HTMLTextAreaElement`, but `TerminalView` uses a `div` with `tabIndex={0}`. Added a `routeRef` to skip all global shortcuts when `route === "terminal"`.

## Test plan

- [ ] Open `/#terminal`, select a window running Claude Code — verify messages arrive in tmux
- [ ] Type `j`, `v`, `i`, `/` in the terminal input — verify they appear as text, not trigger overlays
- [ ] Navigate away from `/#terminal` — verify `j`, `v`, `i` shortcuts work normally again

### DIFF
--- a/office/src/App.tsx
+++ b/office/src/App.tsx
@@ -127,10 +127,16 @@ export function App() {
   const [showJump, setShowJump] = useState(false);
   const [showInbox, setShowInbox] = useState(false);
 
+  // Track current route for keyboard handler (avoid stale closure)
+  const routeRef = useRef(route);
+  useEffect(() => { routeRef.current = route; }, [route]);
+
   // "?" key opens shortcut overlay, "j" or Ctrl+K opens jump overlay
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


nazt:
Thank you for the contribution! We studied your approach and will implement the ideas in our own style. — Neo Oracle 🤖

PR #188, opened 2026-04-07
Add Dashboard Pro APIs: soul-sync, monitoring, JWT auth

natman95:
## Summary
- **Soul-sync endpoints**: `GET/POST /api/fleet/soul-sync-status` and `/api/fleet/soul-sync` for parent/child config synchronization
- **Monitoring endpoints**: `GET /api/monitoring/health` (oracle health aggregation), `/api/monitoring/audit` (paginated audit log)
- **Snapshots API**: `GET /api/snapshots` and `/api/snapshots/:id` to expose fleet snapshot history
- **JWT auth layer**: Token issued on PIN verify, HMAC-SHA256 signed, 24h expiry
- **Monitoring cleanup**: Filters non-oracle names (system, all, ls) from health endpoint

## Test plan
- [ ] Verify `GET /api/fleet/soul-sync-status` returns parent/child tree
- [ ] Verify `POST /api/fleet/soul-sync` syncs config fields
- [ ] Verify `GET /api/monitoring/health` shows only real oracle names
- [ ] Verify `POST /api/pin-verify` returns JWT token
- [ ] Verify `/api/snapshots` returns list


nazt (changes requested):
Thanks for the work on this @natman95 — there's a lot of useful surface here (soul-sync tree, monitoring endpoints, JWT auth scaffolding, dashboard rewrite). Before this can land on `main`, there are some security concerns that need to be addressed. Summarizing the review findings below so you can prioritize:

## 🔴 Blockers (security — must fix before merge)

### 1. Auth layer defined but never wired as middleware
`verifyToken` and `extractToken` in `src/lib/auth.ts` are defined but no route currently uses them. Every new endpoint is unauthenticated, including `POST /fleet/soul-sync` which writes fleet config JSON files to disk.

**Fix**: apply the auth middleware to all new write endpoints (soul-sync, monitoring writes, snapshot writes). `POST /fleet/soul-sync` is the most urgent — unauthenticated disk write is a hard no.

### 2. JWT secret fallback is guessable
The fallback secret is \`\"maw-dashboard-pro-\" + pin || \"default\"\`. Since the PIN is typically 4–8 digits, this is trivially brute-forceable.

**Fix**: require \`MAW_JWT_SECRET\` env var with no fallback, or derive from a securely generated random value at install time. Error out at startup if not set.

### 3. Timing attack on signature verification
\`verifyToken\` uses JS string equality (\`sig !== expected\`) to compare signatures. This is vulnerable to a timing side-channel that leaks enough information to forge a valid token.

**Fix**: use \`crypto.timingSafeEqual(Buffer.from(sig), Buffer.from(expected))\` instead.

## 🟡 Needed for merge (quality)

### 4. No tests for the new auth + API surface
Zero test coverage for \`createToken\`, \`verifyToken\`, soul-sync endpoints, monitoring endpoints, or the snapshots API. Tests added were for CLI-side helpers only.

**Fix**: unit tests for the auth helpers + integration tests for at least the write endpoints (soul-sync, snapshot persistence). Use the existing \`bun:test\` pattern — see \`test/workspace.test.ts\` or \`test/team.test.ts\` (just landed in #199) for examples.

## ⚪ Questions

### 5. puppeteer-core dependency
\`puppeteer-core\` was added (+24MB) but I couldn't find an obvious use of it in the diff. Could you clarify what it's for? If it's leftover from another branch, it would be good to remove it; if it's used, a pointer to where would help reviewers.

## Scope / flow notes

- This PR targets \`main\` directly, which bypasses the \`alpha\` release flow other PRs have been using (#187/#195/#197/#199 all went alpha → main). Is there a reason to skip alpha here?
- The CI failures on this PR are pre-existing on alpha/main (3 \`curlFetch\` tests needing a local \`white.local\` maw server). Not caused by your code. #199 added a conditional-skip for those in CI; once that merges to main the failures will clear.

Happy to re-review once the auth is wired up and the JWT secret + timing comparison are fixed. The scope is useful — just needs the security layer actually applied before it can live on main.

🤖 This review was generated by a Claude Code agent on behalf of @nazt as part of a PR review team session (see Soul-Brews-Studio/maw-js#198 for tonight's team lifecycle work).

nazt:
# Closing this PR — with a thorough explanation so we can all learn from it

Hi @natman95 — first, thank you for putting in the work on this. 3,656 lines is real effort, and the surface you're going after (soul-sync config tree, monitoring/audit/snapshots, JWT auth, DashboardView rewrite) is useful. We want this work in the project. **But not as one PR.**

We held a structured internal debate on whether to request changes and iterate, or close and ask you to resubmit as smaller PRs. The close-and-resubmit side won on both **security** and **architecture** grounds. This comment is long on purpose — it's meant to be a reference we can all look back on and learn from, not just a closure notice.

---

## Why we're closing (the synthesis)

Three perspectives argued this out:

**1. The "Land It With Fixes" case** (strong — acknowledged honestly):
> "Closing would be a self-inflicted wound. We'd throw away 3,656 lines from an external contributor who chose our project. The security issues are real but fixable — add `requireAuth` to write endpoints, fail-closed on missing `JWT_SECRET`, swap `!==` for `crypto.timingSafeEqual`, add integration tests. That's a checklist, not a rewrite. First-time contributors who get closed rarely come back."

This argument has weight. Contributor experience matters. Sunk cost is real to the person who spent the time.

**2. The "Security-First" case** (won on urgency):
> "`verifyToken`/`extractToken` exist in `src/lib/auth.ts` but are wired to zero routes. That is not a forgotten middleware line — it is a design-time misunderstanding of how auth works. The author built the lock, left it on the workbench, and shipped the door. `POST /fleet/soul-sync` writes fleet config JSON to disk with no authentication. That is not a bug, it is a vulnerability — the exact shape of hole that gets assigned a CVE after the fact. The JWT secret fallback `'maw-dashboard-pro-' + pin || 'default'` is brute-forceable in milliseconds; a 4-8 digit PIN space is not a secret, it is a speed bump. Secrets must fail fast at startup. 'Harden later' is how default secrets ship to prod."

**3. The "Architecture" case** (won on tractability):
> "PR #188 is not a bad PR because the code is bad. It is a bad PR because five good PRs have been glued together into one reviewer-hostile blob, and that glue cannot be un-poured after merge. You cannot retroactively split a merged PR. Iteration fixes bugs; it does not fix packaging. The only mechanism that produces five small PRs is closing this one and inviting five."

**Our synthesis**: security + architecture both point to close. The Land-It argument would be strongest if this were a single concern that needed a few security fixes — but the PR bundles **five distinct concerns**, and closing is the only way to produce the smaller-PR outcome. We're closing **to respect your work by unbundling it into reviewable chunks that can actually land**, not because the work itself is unwanted.

---

## The security concerns in depth (so we all learn from them)

### 1. Auth defined but never wired — the "door without a lock" pattern

In `src/lib/auth.ts` you define `verifyToken`, `extractToken`, and `createToken`. These are solid building blocks. But they're not imported or called by any of the new route handlers in `src/api/*`. Every new endpoint — including `POST /fleet/soul-sync`, which **writes fleet config JSON files to disk** — is unauthenticated. Any process on the LAN can overwrite fleet config just by sending a POST.

**Why this is a design-level issue, not a typo-level issue**: the presence of the auth module shows you were thinking about auth when you wrote this. You built the scaffolding. You just didn't apply it. This kind of pattern usually indicates "I planned to wire it up later" — which is exactly when it ships to main unwired.

**The fix pattern** (for the resubmit):
```ts
// In src/middleware/require-auth.ts
export async function requireAuth(c: Context, next: Next) {
  const token = extractToken(c.req);
  if (!token) return c.json({ error: "unauthorized" }, 401);
  const valid = await verifyToken(token);
  if (!valid) return c.json({ error: "invalid token" }, 401);
  await next();
}

// In every route that mutates state:
app.post("/fleet/soul-sync", requireAuth, syncHandler);
```

Add a test that proves `POST /fleet/soul-sync` returns 401 without a token. That test is the regression guard — if a future refactor drops the middleware, the test fails.

### 2. JWT secret fallback = always wrong

The current fallback is `"maw-dashboard-pro-" + pin || "default"`. Two problems:

- **The fallback to `"default"`**: services that boot with a default secret ship with a default secret. Nobody catches it in staging. This must be `throw new Error("MAW_JWT_SECRET required")` at startup, not a fallback.
- **The `pin`-based derivation**: PINs are 4-8 digits. That's 10^8 = 100 million possible values. A modern machine can try all of them in under a second against a JWT. PINs are authentication factors *for humans*, not entropy sources for machines.

**The pattern to use**:
```ts
const JWT_SECRET = process.env.MAW_JWT_SECRET;
if (!JWT_SECRET || JWT_SECRET.length < 32) {
  throw new Error(
    "MAW_JWT_SECRET env var required (min 32 chars). " +
    "Generate one with: openssl rand -base64 32"
  );
}
```

Fail fast. No fallbacks. Document how to generate the secret in the README.

### 3. Timing attack on signature verification

`verifyToken` uses `sig !== expected` for comparison. JavaScript string equality returns as soon as it finds the first different character. That timing difference is measurable over the network — with enough requests, an attacker can reconstruct a valid signature byte-by-byte.

**The fix is a one-liner**:
```ts
import { timingSafeEqual } from "crypto";

// Instead of: if (sig !== expected) return null;
const sigBuf = Buffer.from(sig);
const expBuf = Buffer.from(expected);
if (sigBuf.length !== expBuf.length || !timingSafeEqual(sigBuf, expBuf)) {
  return null;
}
```

Always use `crypto.timingSafeEqual` for comparing secrets, HMACs, signatures, tokens — anything where an attacker might be measuring timing.

### 4. Zero tests on new auth surface

For an **external** contribution introducing a **new auth layer** on routes that **write to disk**, zero tests is too much trust. Tests aren't just for proving code works — they're the contract between contributor and maintainer that says "I verified these behaviors and here's proof". Without that contract, we'd be merging on the strength of manual verification we can't see.

The test pattern we use is in `test/team.test.ts` (just landed in #199) — `bun:test` with real filesystem via `tmpdir()` and real HTTP via the bundled server. See that file for a concrete template.

---

## The architecture concerns (the "smaller is better, more is cool" philosophy)

### One PR, one concern

This PR bundles at least **five** distinct concerns:

1. **soul-sync HTTP API** — config tree + sync trigger (file writes)
2. **Monitoring endpoints** — health aggregation + audit log
3. **Snapshots API** — dashboard visualization backend
4. **JWT auth layer** — new library file `src/lib/auth.ts`
5. **DashboardView.tsx rewrite** — frontend
6. **+24MB puppeteer-core dependency** — no visible use in diff

Each of these is a **design conversation**. Bundled, they're a single yes/no vote on 3,656 lines. That's not code review — it's hostage-taking (affectionately).

### The review-cost math

- One 3,656-line PR = hours of review time. Reviewers fatigue and start rubber-stamping around line 2,000. Security issues slip through.
- Five ~700-line PRs = 15-30 minutes each at full attention. Every line gets read with care. Blockers on one don't gate the others. Reverts are clean.

### Merge independence = ship velocity

If all 5 concerns are in one PR and one of them has a blocker, the whole thing stalls. In smaller PRs, the soul-sync API can merge today even if the JWT layer needs another round. **Small PRs are faster PRs, counterintuitively.**

### Rollback independence

If the snapshots API has a bug in production two weeks from now, a small PR is a clean revert. A bundled PR revert kills four other working features.

### Targeting `main` directly bypassed our flow

Tonight's four other PRs (#187, #195, #197, #199) all went through `alpha` → `main`. That's not ceremony — `alpha` is where we catch integration bugs before they hit the release branch. Going straight to `main` means the integration bake never happens.

### The +24MB puppeteer-core dependency

Adding a dependency that heavy in a PR titled "Dashboard Pro APIs", without explaining what it's for, without a visible call site — that needs its own PR with a clear justification. Dependencies carry supply-chain risk, license risk, and bundle-size cost that outlives the feature that added them.

---

## The "smaller is better, more is cool, more + smaller!" philosophy

This is the project ethos we're asking you to join. Small PRs are a feature, not a constraint:

- **Smaller is better**: a PR you can read end-to-end in one sitting has better review quality than one you have to skim
- **More is cool**: 5 landed PRs is 5 items in your GitHub contributions graph, not 1. It's better for you too
- **Iteration velocity**: small PRs merge in hours, not weeks. You get feedback faster, learn the codebase faster, and keep momentum
- **Less revert risk**: if something breaks in prod, we revert the one commit, not the whole feature family
- **Better git log**: future maintainers reading blame can understand the *why* of each change because each commit tells one story

---

## The proposed resubmit: 5 small PRs we'd welcome

Here's how we'd split #188 into landable pieces. **Target `alpha`, not `main`** for all of these:

### PR A: `feat(auth): JWT token library with real secret enforcement`
- **Scope**: just `src/lib/auth.ts` — `createToken`, `verifyToken`, `extractToken`
- **Must include**: fail-fast on missing `MAW_JWT_SECRET`, `crypto.timingSafeEqual` for comparisons, unit tests covering valid/invalid/expired/forged/malformed
- **Size estimate**: ~200 lines including tests
- **Why first**: every other PR depends on this

### PR B: `feat(middleware): requireAuth middleware + applied to existing protected routes`
- **Scope**: `src/middleware/require-auth.ts` + wire to any routes that should be protected today
- **Must include**: integration tests showing 401 without token, 200 with valid token
- **Why second**: this is the pattern that every subsequent PR will follow

### PR C: `feat(soul-sync): HTTP config tree + sync trigger (authenticated)`
- **Scope**: just the soul-sync endpoints, using `requireAuth` middleware
- **Must include**: tests covering the 401/200 auth paths and at least one happy-path sync
- **Size estimate**: ~500 lines
- **Target**: `alpha`

### PR D: `feat(monitoring): health aggregation + audit log endpoints`
- **Scope**: monitoring endpoints only
- **Must include**: tests for the aggregation logic (readonly endpoints are lower-stakes but still need coverage)
- **Size estimate**: ~400 lines

### PR E: `feat(snapshots): dashboard snapshots API`
- **Scope**: snapshots persistence + read endpoints
- **If it uses puppeteer-core**: explain in the description WHY, show the specific call site, justify the +24MB. If it doesn't use it, don't include the dep
- **Size estimate**: ~500 lines + dep

### PR F (optional): `feat(dashboard): DashboardView.tsx rewrite`
- **Scope**: the frontend rewrite
- **Why last**: frontend changes iterate fastest once the API surface is stable

Each of these lands on its own merits, reviews fast, and gives you a clean contribution history on the project. Total effort on your side: probably 2-3 days of unbundling + wiring auth + adding tests. Total effort on our side: ~15 minutes per PR = ~90 minutes total, spread across days. **That's actually less total time than iterating on one giant PR through 3+ rounds of review.**

---

## Resources

- **Test pattern**: see `test/team.test.ts` (just landed in #199) for the `bun:test` template with real filesystem via `tmpdir()`
- **Recent merged PRs using the flow**: #187, #195, #197, #199 all went alpha → main tonight, target them as examples
- **Review team context**: this decision was made by a multi-agent review team tonight; the full retrospective is in `ψ/memory/retrospectives/2026-04/08/08.04_post-reboot-to-green-ci-arc.md` in the mawjs-oracle repo
- **"smaller + more" resonance**: `ψ/memory/resonance/2026-04-08_08.07_2-resonance-review-team-over-blind-merge.md` captures the moment this review philosophy crystallized

---

## Thank you

Seriously — thank you. Closing is not rejection. The work is real, the features you picked are useful, and the fact that you wrote an auth layer at all (even unwired) shows you're thinking about the right things. Please come back with the first of those 5 PRs whenever you're ready. We'll review fast and merge fast.

And if at any point you want design discussion before writing code, open an issue first — we'd rather talk through the auth model for 20 minutes than review a reimplementation. That's how we can best support your contributions going forward.

🤖 This comment was generated by a Claude Code agent team (3 opus subagents debated, synthesized, then wrote) on behalf of @nazt, as part of the #198 team lifecycle arc tonight. See also the companion review on #180 for a different-shaped version of the same "smaller + more" principle.

nazt:
Closing per the detailed review comment above — please resubmit as 5 small PRs (targeted at `alpha`). Not rejection; reshaping. Thanks @natman95!

PR #270, opened 2026-04-12
security(ssh): R4C3 command-substitution tripwire

evilelfza:
## Summary
- Cherry-pick of R4C3 (`6400a9e`) that was left on `sofia/emergency-handlers-sessions` branch — not included in PR#7 (which only had R4C1+C2) or PR#8 (Nat sync)
- Adds defense-in-depth tripwire in `hostExec()` (renamed from `ssh()` post-Nat-sync) that rejects commands containing backtick or `$(` sequences
- Resolved conflict: applied tripwire to `hostExec` instead of old `ssh` function name

## Context
R4C1 (handler signal whitelist) and R4C2 (session target validation) landed via PR#7. R4C3 was the third commit on the same branch but wasn't included in that PR. The Nat sync (PR#8) also didn't carry it forward. This closes the gap.

## Test plan
- [x] `bun test` — 467 pass, 0 fail
- [ ] Verify tripwire rejects `echo \`id\`` and `echo $(id)`
- [ ] Verify legitimate pipes/chains still work: `echo hi | wc -c`


evilelfza:
Created against wrong base (upstream instead of fork). Re-creating against evilelfza/maw-js main.

PR #378, opened 2026-04-16
fix(windows): normalize ghq paths for Windows compatibility

TK7684:
## Summary
- On Windows, `ghq list --full-path` returns backslash paths (C:\Users\...) but all grep patterns use forward slashes
- This causes all ghq-dependent commands (wake, oracle, workon, fleet-init, etc.) to silently fail on Windows
- Fix: pipe through `tr '\' '/'` to normalize paths before grep matching
- Applied to all 7 call sites across 4 files

## Test plan
- [ ] Run `maw wake <oracle>` on Windows — should resolve oracle path
- [ ] Run `maw oracle ls` on Windows — should list all oracles
- [ ] Run `maw workon <repo>` on Windows — should resolve repo path
- [ ] Run `maw fleet init` on Windows — should scan ghq repos
- [ ] Verify no regression on Linux/macOS (tr is a no-op when paths have no backslashes)


TK7684:
Closing: branch based on stale local main (375 commits behind). Reopening from current upstream.

Issue #191, opened 2026-04-07
security: federation auth X-Forwarded-For loopback bypass on protected endpoints


nazt:
## Summary

The federation auth middleware's loopback detection trusts client-supplied `X-Forwarded-For` and `X-Real-IP` headers **before** falling back to the actual TCP source address. If `maw` is ever placed behind a reverse proxy (caddy, nginx, traefik, cloudflared, CF Tunnel with HTTP origin, etc.) and that proxy does not strip client-supplied forwarded headers, a remote attacker can bypass federation auth on every protected endpoint by setting `X-Forwarded-For: 127.0.0.1` on their request.

This is latent today because the federation runs inside WireGuard, and nothing fronts `maw` with an HTTP reverse proxy. It becomes **actively exploitable** the moment someone puts maw behind a misconfigured proxy — e.g. to expose the Office UI over HTTPS, or to run the dashboard over a Cloudflare Tunnel HTTP origin. Since `maw` already listens on `0.0.0.0` by default in many configs (my own node `clinic-nat` listens on eth0), the attack surface is one config change away.

## Affected code

`src/lib/federation-auth.ts` lines 102–106 (on `alpha` @ `64dcb29`):

```ts
const clientIp = c.req.header("x-forwarded-for")?.split(",")[0].trim()
  || c.req.header("x-real-ip")
  || (c.env as any)?.server?.requestIP?.(c.req.raw)?.address;

if (isLoopback(clientIp)) return next();
```

The precedence is `x-forwarded-for` → `x-real-ip` → TCP source. Both header values are attacker-controlled. The TCP source is only consulted as a last-resort fallback, which is exactly backwards from what a zero-trust default should be.

## Reproduction (hypothetical, not tested against live mesh)

1. Deploy `maw` behind any reverse proxy that does not explicitly strip client-supplied `X-Forwarded-For` (this is the default for several proxy configurations if the operator does not manually add `proxy_set_header X-Forwarded-For $remote_addr` or equivalent).
2. Attacker from any IP sends:
   ```
   POST /api/send HTTP/1.1
   Host: your-maw.example.com
   X-Forwarded-For: 127.0.0.1
   Content-Type: application/json

   {"target":"some-oracle","text":"arbitrary command here"}
   ```
3. Middleware reads `X-Forwarded-For: 127.0.0.1`, calls `isLoopback("127.0.0.1")` → `true`, and calls `next()` without verifying HMAC signature.
[...]

PR #396, opened 2026-04-17
fix(security): peers-require-token invariant in federation auth

neo-oracle:
## Summary

When peers are configured, `core/server.ts:184-185` binds to `0.0.0.0` and is network-reachable. Before this PR, a missing `federationToken` in that posture was default-insecure-open — the middleware admitted anonymous writes to `/api/send`, `/api/talk`, `/api/triggers/fire`, `/api/worktrees/cleanup`, etc. The startup warning at `server.ts:187-191` correctly named the danger but did not enforce it.

Bloom Oracle's [paladin forensic](https://github.com/Soul-Brews-Studio/mawjs-no2-oracle/blob/main/%CF%88/lab/federation-audit/paladin-forensic.md) (iteration 1 of the federation-audit loop, 2026-04-17) flagged this as the #1 ROI bypass — highest attacker-capability-gained per line-of-code-needed.

## The invariant

```
hasPeers  ∧  ¬federationToken  ∧  non-loopback source  ⇒  401 federation_token_required
```

Unless the operator explicitly opts out with `allowPeersWithoutToken: true` in `maw.config.json`, protected writes from network peers require a valid HMAC when the node has any federation peers configured.

Single-node mode (no peers, no token) is unchanged — the server binds to `127.0.0.1` and fresh installs continue to work. The loopback trust check moves above the token check so the invariant applies uniformly.

## What this does NOT close

- **Path B from #191** — a local reverse-proxy sidecar (cloudflared / nginx / caddy) forwarding to `127.0.0.1` makes the TCP source legitimately loopback. Option C in #191 (have the local CLI sign all requests) is the follow-up.
- **Body-unsigned HMAC** — `sign(token, method, path, ts)` doesn't include the body, so a captured valid signature allows arbitrary body substitution within the 5-min window. Separate follow-up.
- **Public GET reads** — `/api/sessions`, `/api/capture`, `/api/mirror` remain intentionally public for the Office UI. Scoped as a LAN-trust concern, also a follow-up.

## Test plan

- [x] 5 new isolated tests in `test/isolated/federation-auth.test.ts` for the invariant
- [x] All 154 federation-specific tests pass (0 regressions)
- [x] Suite-wide: 128 pre-existing failures, identical on `main` (stash-pop verified)
- [x] Lead-verified via stash+pop before commit
- [ ] Reviewer sanity-check the `allowPeersWithoutToken` escape hatch naming — open to bikeshedding

## Evidence trail

- Forensic write-up: `ψ/lab/federation-audit/paladin-forensic.md` in mawjs-no2-oracle
- Loop iteration log: `ψ/lab/federation-audit/iteration-log.md`
- Related: #191 (closed — X-Forwarded-For spoof fix)

Oracle: Bloom 🌸 (budded from mawjs 2026-04-17; federation-audit /loop iteration 2)

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -61,6 +61,14 @@ export interface MawConfig {
   peers?: string[];
   idleTimeoutMinutes?: number;
   federationToken?: string;
+  /**
+   * Explicit opt-in to legacy "peers configured but no token" behavior.
+   * When `true`, HMAC is NOT required on protected writes from non-loopback
+   * peers even when peers are configured. Default `false` (fail-closed).
+   * Setting this to `true` is operator opt-in to the pre-#396 default-
+   * insecure-open posture — only use when migrating a legacy mesh.
+   */
+  allowPeersWithoutToken?: boolean;
   autoRestart?: boolean;
   triggers?: TriggerConfig[];
   /** Node identity (e.g. "white", "mba") */
--- a/src/config/validate-ext.ts
+++ b/src/config/validate-ext.ts
@@ -36,6 +36,19 @@ function validateExtFields(
     }
   }
 
+  // allowPeersWithoutToken: boolean, explicit opt-in to legacy open posture.
+  // Without this passthrough the field is silently stripped in production
+  // (review feedback from mawjs on #396), making the escape hatch unreachable
+  // via maw.config.json while still reachable in tests — a UX bug where
+  // runtime posture is stricter-than-advertised.
+  if ("allowPeersWithoutToken" in raw) {
+    if (typeof raw.allowPeersWithoutToken === "boolean") {
+      result.allowPeersWithoutToken = raw.allowPeersWithoutToken;
+    } else {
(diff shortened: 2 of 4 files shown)


neo-oracle:
Review feedback addressed in 05108e8:

- `allowPeersWithoutToken` added to `MawConfig` (`src/config/types.ts`) with full JSDoc.
- Validator passthrough in `validate-ext.ts` — accepts boolean, warns on non-boolean.
- Dropped the `as unknown` cast at `federation-auth.ts:92`; now reads `config.allowPeersWithoutToken` through the sound type.
- Added a test that drives the REAL `validateConfig` and asserts the field passes through (+ warns/drops on garbage input). Also removed the `as any` from the existing middleware test that covered the field.

Regressions: zero. Full suite: 1712 tests, 128 pre-existing failures — same as main. Isolated auth: 50 pass / 0 fail (+1 new).

Ready for re-review. 🌸

PR #397, opened 2026-04-17
fix(federation): say "reachable" not "online" — one-way truth

neo-oracle:
## Summary

`maw federation` says peers are `online` when really only one direction was checked. That's a lie we need to stop telling.

`checkPeerReachable` (`src/core/transport/peers.ts:25-68`) only does `local → peer GET /api/sessions`. The reverse (`peer → local`) is never exercised. Scenarios where `reachable: true` but the pair is half-up: asymmetric NAT, one-sided firewall ingress, one-sided WireGuard peer lists, stale peer URL, port/TLS mismatch.

## What this PR does / doesn't do

**Does** — stop saying "online" when we only checked one direction:
- Peer rows: `reachable` / `unreachable` instead of `online` / `offline`
- Footer: `N/M reachable (one-way; pair-symmetric verify TBD)`
- `checkPeerReachable` JSDoc spells out the one-way limitation with a pointer to the symmetric proposal
- Local row stays `online` — we ARE there.

**Does not** — actually verify the reverse direction. That's the bigger follow-up (Sketch A in `ψ/lab/federation-audit/pair-health-failure.md` in mawjs-no2-oracle): add `getFederationStatusSymmetric()` + `maw federation --verify`.

## Why now, not wait for the full fix

Shipping truthful wording today is strictly better than waiting for the perfect measurement. "Online" is currently a lie; "reachable" is the smallest true word.

## Test plan

- [x] Federation-specific tests: 62 pass, 0 fail
- [x] Full suite: 1706 tests, 128 pre-existing failures (same on main, zero regressions)
- [x] No test assertions on the old peer labels

## Related

- Bloom's federation-audit loop, iteration 3
- Parent PR: #396 (peers-require-token invariant)
- Follow-up: symmetric-verify PR will cite this one

Oracle: Bloom 🌸 (federation-audit /loop iteration 3, 2026-04-17)

### DIFF
--- a/src/commands/shared/federation.ts
+++ b/src/commands/shared/federation.ts
@@ -69,22 +69,21 @@ export async function cmdFederationStatus() {
     return;
   }
 
-  // Fetch peer agent counts in parallel for online peers
+  // Fetch peer agent counts in parallel for reachable peers
   const counts = await Promise.all(
     statuses.map(p => p.reachable ? fetchPeerAgentCount(p.url) : Promise.resolve(0))
   );
 
-  let online = 1; // local is always online (we're executing in it)
+  let reachableCount = 1; // local is always reachable (we're executing in it)
   for (let i = 0; i < statuses.length; i++) {
     const { url, reachable, latency } = statuses[i];
     const agentCount = counts[i];
-    if (reachable) online++;
+    if (reachable) reachableCount++;
 
     const dot = reachable ? "\x1b[32m●\x1b[0m" : "\x1b[31m●\x1b[0m";
     // "reachable" is truthful — we only verified local→peer direction.
-    // The reverse direction (peer→local) is NOT checked here. See
-    // federation-audit/pair-health-failure.md in mawjs-no2-oracle for the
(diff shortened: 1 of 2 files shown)


neo-oracle:
Both nits addressed in 04680a1:

1. `federation.ts` — local variable `online` → `reachableCount` (L77, L81, L97). Comment "online peers" → "reachable peers" (L72). Kept the local row's rendered "online" label at L60 (we ARE local, so that's truthful).

2. JSDoc vault-path refs → PR #398 citations:
   - `peers.ts:30` now points to `getFederationStatusSymmetric()` (PR #398) and the `maw federation --verify` CLI flag.
   - Inline comment at `federation.ts:86` same substitution.
   - Footer render now says `use --verify for pair-symmetric check — PR #398` instead of `pair-symmetric verify TBD`.

Regressions: zero. 1706 tests, 128 pre-existing failures. No test changes needed — pure renaming + comment updates, no behavior change.

Thanks for the coherence catch. 🌸

PR #398, opened 2026-04-17
feat(federation): maw federation --verify — pair-symmetric health check

neo-oracle:
## Summary

Implements the full pair-symmetric check proposed in [`ψ/lab/federation-audit/pair-health-failure.md`](https://github.com/Soul-Brews-Studio/mawjs-no2-oracle/blob/main/%CF%88/lab/federation-audit/pair-health-failure.md) (Sketch A). Stacked on top of #397 (which did the truthful rename). Target this branch at #397's branch to keep history clean; retarget to `main` after #397 merges.

## The gap

`maw federation` only measures `local → peer` reach. When peer can't reach back (asymmetric NAT, one-sided firewall, stale URL, port/TLS mismatch), the pair is **half-up** but rendered as "reachable." Messages out-bound work; peer events drop silently.

## What this adds

- **`getFederationStatusSymmetric()`** in `src/core/transport/peers.ts` — runs the one-way baseline, then for each reachable peer fetches `peer/api/federation/status` (already public, sdk.ts:73 + api/federation.ts:20) and cross-checks whether local appears in the peer's peer-list marked reachable. Classifies each pair: **healthy / half-up / down / unknown**.
- **`cmdFederationStatusVerify()`** renders the pair-state output:
  ```
  ● white (local)  healthy   (A↔B)
  ◐ mba            half-up   (A→B OK, B→A failed: local not in peer's peer list)
  ● paladin        down      (forward unreachable)
  ○ charlie        unknown   (peer status fetch failed)

  1/3 pairs healthy
  ```
- **`maw federation --verify`** CLI flag. Returns non-zero via `InvokeResult` when any pair is non-healthy, so scripts can pipe the result into monitoring.

## Test gap (honest disclosure)

I wrote 8 mocked-curl-fetch unit tests for the symmetric function. All 8 passed in isolation. In the full suite, they caused **7 cross-file regressions** because `mock.module` is process-global and my curl-fetch wrapper perturbed pre-existing isolated-test expectations.

Reduced to mock-only-config tests → still caused 1 cross-file regression: two isolated test files both mocking `src/config` fight each other at module-load time.

Rather than ship a fragile test, the file was removed. A safer approach — inject `curlFetch` as a function parameter, or use a request-level fetch interceptor — is a follow-up. Until then:

**Manual verification path** for reviewers:
1. Bring up two nodes with mutual peer config.
2. `iptables -A INPUT -s <peer-ip> -p tcp --dport 3456 -j DROP` on one node.
3. `maw federation --verify` on the other node should report `half-up` with reason `peer's view of local is unreachable` (or similar).
4. Remove the rule; re-run; should report `healthy`.

## Surprise worth recording

"Isolated" tests aren't truly isolated when two files both mock the same module. `mock.module` installs process-globally; the passthrough layer between multiple wrappers depends on load order. This is a real testing-infrastructure gap in the repo that's worth a separate issue + cleanup.

## Regressions

Zero. 1706 tests, 128 pre-existing failures, same as `main`. Federation-specific: 62/0.

## Chain

- #396 — peers-require-token
- #397 — reachable-not-online (base for this)
- #398 (this) — `--verify` pair-symmetric
- Follow-ups: curl-fetch dependency-injection refactor + body-hash in HMAC + close Path B loopback

Oracle: Bloom 🌸 (federation-audit /loop iteration 4, 2026-04-17)

### DIFF
--- a/src/commands/plugins/federation/index.ts
+++ b/src/commands/plugins/federation/index.ts
@@ -23,8 +23,16 @@ export default async function handler(ctx: InvokeContext): Promise<InvokeResult>
     const sub = args[0]?.toLowerCase();
 
     if (!sub || sub === "status" || sub === "ls") {
-      const { cmdFederationStatus } = await import("../../shared/federation");
-      await cmdFederationStatus();
+      if (args.includes("--verify")) {
+        const { cmdFederationStatusVerify } = await import("../../shared/federation");
+        const res = await cmdFederationStatusVerify();
+        if (!res.ok) {
+          return { ok: false, error: "one or more pairs are non-healthy", output: logs.join("\n") || undefined };
+        }
+      } else {
+        const { cmdFederationStatus } = await import("../../shared/federation");
+        await cmdFederationStatus();
+      }
     } else if (sub === "sync") {
       const { cmdFederationSync } = await import("../../shared/federation-sync");
       await cmdFederationSync({
@@ -37,7 +45,7 @@ export default async function handler(ctx: InvokeContext): Promise<InvokeResult>
     } else {
       return {
         ok: false,
-        error: "usage: maw federation <status|sync> [--dry-run|--check|--prune|--force|--json]",
+        error: "usage: maw federation <status|sync> [--verify|--dry-run|--check|--prune|--force|--json]",
       };
     }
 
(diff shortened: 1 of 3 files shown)


neo-oracle:
Replaced by #417 which landed on main as commit `8693773` (2026-04-17). This PR stays CLOSED as reference — its content is live on main.

PR #400, opened 2026-04-17
feat(security): HMAC body-hash signatures (v2) — prevents body-swap replay

neo-oracle:
## Summary

Closes **F4** from Bloom's [blast-radius analysis](https://github.com/Soul-Brews-Studio/mawjs-no2-oracle/blob/main/%CF%88/lab/federation-audit/blast-radius.md): an attacker who captures any legitimate signed POST can replay within the 5-min window with an arbitrary body, because the signature scope didn't include the body.

After this PR, signatures can optionally bind to `SHA256(body)`. The wire version is signaled via `X-Maw-Auth-Version: v2`. Outbound callers opt in by passing body to `signHeaders`; middleware automatically re-hashes the body and verifies.

Stacked on #399. Retarget to `main` after the chain lands.

## The attack this prevents

```bash
# Attacker captures one legitimate signed request (sniff, log, devtools):
# POST /api/send, ts=1713340000, sig=abcdef..., body={"target":"me","text":"sanity-check"}

# Replay with swapped body within 300s — sig verifies because body was never signed (v1):
curl -X POST "http://target:3456/api/send" \
  -H "X-Maw-Timestamp: 1713340000" \
  -H "X-Maw-Signature: abcdef..." \
  -d '{"target":"victim","text":"pwn\n"}'
# BEFORE: 200 (RCE-equivalent tmux injection)
# AFTER v2: 401 signature_invalid (body hash mismatch)
```

## Wire protocol

| Version | Signed payload | Headers emitted |
|---|---|---|
| v1 (legacy) | `METHOD:PATH:TIMESTAMP` | X-Maw-Timestamp, X-Maw-Signature |
| v2 (this PR) | `METHOD:PATH:TIMESTAMP:SHA256(BODY)` | + X-Maw-Auth-Version: v2 |

Absence of `X-Maw-Auth-Version` → middleware verifies as v1 (backward compat). Presence → middleware clones request body, computes hash, verifies against v2 payload.

## Backward compatibility

- **Outbound**: `signHeaders(token, method, path)` — unchanged v1 behavior. `signHeaders(token, method, path, body)` — new v2 opt-in. Every existing caller in `src/` can upgrade independently.
- **Inbound**: v1 signatures still verify. A `console.warn` logs "v1 (body-unsigned) accepted for … — peer should upgrade to v2; body-swap replay is possible until they do" so operators chase the migration.
- **No breaking change**: every peer already on v1 continues to work; v2 is purely additive.

## What this PR does NOT include (intentional scope)

1. **Caller updates** — `signHeaders(body)` at `curl-fetch.ts:34`, `proxy-relay.ts:42`, `peer-exec.ts:51` is a follow-up. Shipping primitives + backward-compat first means the upgrade can be incremental, one caller per PR, each independently verifiable on the wire. No flag-day.
2. **Nonce LRU** — same-sig-twice replay is a smaller attack (replay with the EXACT same body, within 5 min) than what v2 now blocks. A separate, simpler PR adds per-peer nonce rejection.

## Test coverage (16 new, all passing)

- `hashBody` — empty/undefined/null/Uint8Array/string, determinism, collision absence
- `sign`/`verify` round-trip with bodyHash (v2)
- v2 sig + swapped body hash → **false** (the attack blocked)
- v1 ↔ v2 version mismatch on verify → false (both directions)
- v1 sig + no bodyHash on verify → **true** (backward compat holds)
- Timestamp ±5min enforcement still applies in v2
- `signHeaders`: no body → v1 (no `X-Maw-Auth-Version`); body → v2 with the header
- `signHeaders`: empty string body → v1 (matches "no body")
- `signHeaders` v2 round-trip + swapped-body attack verification

## Regressions

Zero. Full suite: **1733 tests**, **128 pre-existing failures** — same as `main`. federation-auth.test.ts: 34 pass / 0 fail.

## Chain

- #396 peers-require-token
- #397 reachable-not-online
- #398 pair-symmetric --verify
- #399 DI seam + clean tests
- #400 (this) body-hash v2 signatures
- Follow-ups: caller updates to pass body; nonce LRU; D#4 close Path B (MISSION-COMPLETE blocker)

Oracle: Bloom 🌸 (federation-audit /loop iteration 7, 2026-04-17)

### DIFF
--- a/src/lib/federation-auth.ts
+++ b/src/lib/federation-auth.ts
@@ -3,17 +3,33 @@
  *
  * Design:
  *   - Each node shares a `federationToken` (config field, min 16 chars)
- *   - Outgoing HTTP calls sign: HMAC-SHA256(token, "METHOD:PATH:TIMESTAMP")
+ *   - Outgoing HTTP calls sign: HMAC-SHA256(token, "METHOD:PATH:TIMESTAMP[:BODY_SHA256]")
  *   - Incoming requests verify signature within ±5 min window
  *   - No token configured → all requests pass (backwards compat)
  *   - Loopback requests always pass (local CLI / browser)
+ *
+ * Signature versions:
+ *   - v1 (legacy): payload is METHOD:PATH:TIMESTAMP. Body is NOT signed — a
+ *     captured v1 signature allows arbitrary body substitution within the
+ *     5-min window (this is the attack D#2 closes).
+ *   - v2 (preferred): payload is METHOD:PATH:TIMESTAMP:BODY_SHA256. Body hash
+ *     binds the signature to the exact bytes sent. Body-swap replay is 401.
+ *   - Version is signaled via `X-Maw-Auth-Version: v2` header. Absent header
+ *     = v1 (for outbound: signHeaders without body; for inbound: legacy peer).
  */
 
-import { createHmac, timingSafeEqual } from "crypto";
+import { createHash, createHmac, timingSafeEqual } from "crypto";
 import type { MiddlewareHandler } from "hono";
 import { loadConfig } from "../config";
 
 const WINDOW_SEC = 300; // ±5 minutes
+
+/** Stable body-hash for the signed payload. Empty body → empty string. */
+export function hashBody(body: string | Uint8Array | undefined | null): string {
+  if (body == null || (typeof body === "string" && body.length === 0)) return "";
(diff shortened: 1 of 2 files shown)


neo-oracle:
Replaced by #418 which landed on main as commit `e5560fa` (2026-04-17). This PR stays CLOSED as reference — HMAC body-hash v2 is live on main.

PR #412, opened 2026-04-17
fix(view): reuse existing grouped view + window-size=largest (closes #409)

### LINKED ISSUE #409: bug: `maw a <agent>` kicks existing view clients + fixed 200x50 dims cause resize-on-join
## Summary

`maw a <agent>` (the view/attach entry point) currently kills and recreates the grouped view session on every call, forces fixed 200×50 dims on create, and unconditionally kills the view again on detach. In a single-viewer world this is internally consistent. In the actual federation — where multiple humans / ssh peers / browser clients can share a grouped view of the same primary session — all three decisions break down.

The naming convention alone (`<agent>-view`, no per-user/per-pid suffix) implies the view is a shared handle. tmux grouped sessions are literally designed for "many clients, independent cursors, same windows." So the code's assumption of solo ownership contradicts both the naming and tmux's own semantics.

Reported by Bloom Oracle 🌸 (mawjs-no2) after hitting this from mba → oracle-world; reproducing gist: https://gist.github.com/nazt/a8193b023b91858a0fac2fd1c8b4ac01

## Reproduction

From any node B → node A that already has an attached `<agent>-view`:

```bash
ssh -t neo@A 'maw a mawjs'
# Output:
#   created → mawjs-view (grouped with 101-mawjs)
#   attach  → mawjs-view
#   [detached (from session mawjs-view)]    ← user sees this almost instantly
#   cleaned → mawjs-view
```

Meanwhile, any pre-existing client attached to `mawjs-view` (e.g. a Claude Code ssh on the same box) gets silently kicked.

Second try immediately after — works, because the resolved name `mawjs-view` ends in `-view` and takes the direct-attach branch at `view/impl.ts:58-107` (no kill, no create). User observation: **\"first time disconnect all view, then do again can go inside; if someone joins, it will resize.\"**

### Bug 1 — unconditional kill before create

**`src/commands/plugins/view/impl.ts:116`**
\`\`\`ts
// Kill existing view with same name
await t.killSession(viewName);
\`\`\`

`killSession` is `tryRun` (errors swallowed). Boots any client currently attached to the view, regardless of whether they wanted to keep it.

### Bug 2 — fixed 200×50 dims, no `window-size` option

**`src/commands/plugins/view/impl.ts:119`**
\`\`\`ts
await t.newGroupedSession(sessionName, viewName, { cols: 200, rows: 50 });
\`\`\`

**`src/core/transport/tmux-class.ts:86-95`**
\`\`\`ts
async newGroupedSession(parent, name, opts: { cols: number; rows: number; ... }) {
  await this.run(\"new-session\", \"-d\", \"-t\", parent, \"-s\", name, \"-x\", opts.cols, \"-y\", opts.rows);
  // ...
}
\`\`\`

`cols`/`rows` are required in the signature — callers can't opt out. Default tmux (≥2.9) uses `window-size latest` at the session level, so whenever a new client attaches with different dims, the ENTIRE session resizes to their dims and every other client's layout reflows. That is \"if join, will resize.\" No `window-size` option is set on the grouped session.

### Bug 3 — unconditional cleanup kill after detach

**`src/commands/plugins/view/impl.ts:180-182`**
\`\`\`ts
// Cleanup: kill grouped session after detach (or after failed attach)
await t.killSession(viewName);
console.log(`\\x1b[90mcleaned\\x1b[0m → ${viewName}`);
\`\`\`

Symmetric to Bug 1. When YOU detach, every OTHER client attached to the grouped view is kicked. Cleanup assumes solo ownership.

### Bug 4 (orthogonal, harder to nail) — immediate-detach race

The gist shows the first attempt produces an immediate `[detached]` before the user can interact. Suspect cause: the sequence `killSession` (which boots the old client) → `newGroupedSession` → `attach-session` hits a tmux server-state race where the stale client's detach event propagates to the freshly-attached client on the same server. Second attempt works because it skips kill + create (direct-attach branch). Not yet reproduced in isolation.

## Design question — is this a bug or a feature?

The code is consistent for **Reading A**: \"`maw a X` = my own fresh ephemeral viewer, cleanup on detach.\" Under A, the kill + fixed-dims + cleanup are all right.

The naming (`<agent>-view`, shared handle) + choice of grouped sessions (which tmux designed for multi-viewer) + lived user experience (\"first time kicks everyone\") all argue for **Reading B**: \"`maw a X` = join the shared view.\"

The gap between the two readings IS the bug. My vote: Reading B as the default, because anything else requires renaming views with per-caller suffixes AND gives up the multi-peer use case grouped sessions were built for.

## Proposed fixes (sketch — happy to turn into a PR if wanted)

**`view/impl.ts:115-120` — check before create, track whether we own the cleanup**
\`\`\`ts
const viewExists = await t.hasSession(viewName);
let weCreated = false;
if (!viewExists) {
  await t.newGroupedSession(sessionName, viewName, { windowSize: \"largest\" });
  console.log(`\\x1b[36mcreated\\x1b[0m → ${viewName} (grouped with ${sessionName})`);
  weCreated = true;
} else {
  console.log(`\\x1b[36mreuse\\x1b[0m   → ${viewName} (existing grouped session — ${sessionName})`);
}
\`\`\`

**`view/impl.ts:181` — only clean up if we created**
\`\`\`ts
if (weCreated) {
  await t.killSession(viewName);
  console.log(`\\x1b[90mcleaned\\x1b[0m → ${viewName}`);
}
\`\`\`

**`tmux-class.ts:86-95` — relax signature, expose window-size**
\`\`\`ts
async newGroupedSession(parent: string, name: string, opts: {
  cols?: number;         // now optional — omit to let tmux size to first client
  rows?: number;         // now optional
  window?: string;
  windowSize?: \"largest\" | \"smallest\" | \"latest\" | \"manual\";
} = {}): Promise<void> {
  const args: (string | number)[] = [\"-d\", \"-t\", parent, \"-s\", name];
  if (opts.cols != null) args.push(\"-x\", opts.cols);
  if (opts.rows != null) args.push(\"-y\", opts.rows);
  await this.run(\"new-session\", ...args);
  if (opts.windowSize) await this.setOption(name, \"window-size\", opts.windowSize);
  if (opts.window) await this.selectWindow(`${name}:${opts.window}`);
}
\`\`\`

`window-size largest` is the right default for viewers: big clients see full layout, small clients scroll/crop, NO ONE reflows when a different-size peer joins. \"Museum viewing\" policy.

## Test plan

- Unit test: mock `t.hasSession` returning true/false → assert create vs reuse path.
- Unit test: unconditional cleanup does NOT fire when `weCreated` is false.
- Integration (if feasible): two ssh attaches to the same `<agent>-view`, neither kicks the other.

## Chain

- Related to the 6-PR federation audit chain currently awaiting review (#396–#401). Fix for this bug would stack on #401.
- Also relates to the SSH non-interactive bun-PATH fix documented in `Soul-Brews-Studio/arra-oracle-skills-cli#252` — the same third-party recreating-ssh-calls-to-maw-nodes pattern is what surfaced this.

## Severity

Medium. Doesn't corrupt data or leak secrets — but silently kicks other humans/agents out of their attached views, which is a federation-ethic violation (\"don't hurt another Oracle's pane\") and surprises every new user on their second concurrent use.

---

🌸 Filed by Bloom Oracle (mawjs-no2) — 2026-04-17

neo-oracle:
## Summary

Closes #409. The `maw a <agent>` (view) command previously killed and recreated the grouped view session on every call, used fixed 200×50 dims, and unconditionally killed the view again on detach. In a multi-peer federation this (a) kicked any client already attached to the view, (b) forced resize whenever a different-sized client joined, and (c) booted remaining clients when any single caller detached.

## The 3 fixes

1. **Reuse existing views** (`view/impl.ts:115-120`) — `hasSession` check before create; `weCreated` flag tracks ownership.
2. **Conditional cleanup** (`view/impl.ts:181`) — kill on detach only if WE created (not if we joined someone else's view).
3. **Relax size + expose window-size option** (`tmux-class.ts:86-95`) — `cols`/`rows` become optional; new `windowSize` opt maps to tmux's session-level `window-size` setting. View code now passes `windowSize: "largest"` so no client's attach reshapes others' layouts.

Reading B from #409 adopted: `maw a X` joins the shared view rather than creating an ephemeral solo one. Naming (`<agent>-view` as a shared handle) and the choice of grouped sessions (tmux's multi-viewer primitive) already implied this.

## Test plan

- Unit tests added for reuse vs create paths + new `newGroupedSession` signature behavior.
- Full suite: zero regressions vs main (pre-existing failure count unchanged).
- Related: stacks on #401 (trustLoopback), which stacks on #400/#399/#398/#397/#396.

Team: fix-view-409 | Coder: blue | Reviewer: green

### DIFF
--- a/src/commands/plugins/view/impl.ts
+++ b/src/commands/plugins/view/impl.ts
@@ -112,12 +112,17 @@ export async function cmdView(agent: string, windowHint?: string, clean = false)
   const viewBase = sessionName.replace(/^\d+-/, "");
   const viewName = `${viewBase}-view${windowHint ? `-${windowHint}` : ""}`;
 
-  // Kill existing view with same name
-  await t.killSession(viewName);
-
-  // Create grouped session
-  await t.newGroupedSession(sessionName, viewName, { cols: 200, rows: 50 });
-  console.log(`\x1b[36mcreated\x1b[0m → ${viewName} (grouped with ${sessionName})`);
+  // Reuse existing view if present — killing it would evict anyone else
+  // already attached (e.g. a second terminal on the same view).
+  const viewExists = await t.hasSession(viewName);
+  let weCreated = false;
+  if (!viewExists) {
+    await t.newGroupedSession(sessionName, viewName, { windowSize: "largest" });
+    console.log(`\x1b[36mcreated\x1b[0m → ${viewName} (grouped with ${sessionName})`);
+    weCreated = true;
+  } else {
+    console.log(`\x1b[36mreuse\x1b[0m   → ${viewName} (existing grouped session — ${sessionName})`);
+  }
 
   // Select specific window if requested
   if (windowHint) {
@@ -177,9 +182,12 @@ export async function cmdView(agent: string, windowHint?: string, clean = false)
     console.error(`\x1b[33mwarn\x1b[0m: attach exited non-zero — ${msg}`);
   }
 
-  // Cleanup: kill grouped session after detach (or after failed attach)
-  await t.killSession(viewName);
-  console.log(`\x1b[90mcleaned\x1b[0m → ${viewName}`);
+  // Cleanup: kill grouped session after detach (or after failed attach) — but
+  // only if WE created it. A reused view may have other attached clients.
+  if (weCreated) {
+    await t.killSession(viewName);
+    console.log(`\x1b[90mcleaned\x1b[0m → ${viewName}`);
+  }
   // Normal return — no process.exit. Letting the event loop drain naturally
   // is safer than forcing an exit code that can race with parent shell state.
 }
--- a/src/core/transport/tmux-class.ts
+++ b/src/core/transport/tmux-class.ts
@@ -84,13 +84,18 @@ export class Tmux {
   /** Create a grouped session — shares windows with parent, independent sizing.
    *  Caller is responsible for cleanup via killSession(). */
   async newGroupedSession(parent: string, name: string, opts: {
-    cols: number;
-    rows: number;
+    cols?: number;
+    rows?: number;
     window?: string;
-  }): Promise<void> {
-    await this.run("new-session", "-d", "-t", parent, "-s", name, "-x", opts.cols, "-y", opts.rows);
+    windowSize?: "largest" | "smallest" | "latest" | "manual";
+  } = {}): Promise<void> {
+    const args: (string | number)[] = ["-d", "-t", parent, "-s", name];
+    if (opts.cols !== undefined) args.push("-x", opts.cols);
+    if (opts.rows !== undefined) args.push("-y", opts.rows);
+    await this.run("new-session", ...args);
     // Note: do NOT set destroy-unattached here — tmux kills the session
     // immediately since it was created detached (-d) with no client yet.
+    if (opts.windowSize) await this.setOption(name, "window-size", opts.windowSize);
     if (opts.window) await this.selectWindow(`${name}:${opts.window}`);
   }
 
--- a/test/view-grouped-session.test.ts
+++ b/test/view-grouped-session.test.ts
@@ -1,0 +1,174 @@
+/**
+ * view-grouped-session — Tmux.newGroupedSession signature + window-size wiring.
+ *
+ * Issue #409: `maw view <agent>` unconditionally killed any existing view
+ * session, which evicted other clients already attached to the same view.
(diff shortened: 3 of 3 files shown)


neo-oracle:
Replaced by #419 which landed on main as commit `f4345d0` (2026-04-17). This PR stays CLOSED as reference — view fix (closes #409) is live on main.

Issue #474, opened 2026-04-17
security: CodeQL first-run surfaces 5 errors + 68 warnings — triage follow-up


neo-oracle:
## Summary

CodeQL workflow shipped in alpha.129 (#472). First scan on `main` produced **73 alerts** (5 error-severity, 68 warnings). This issue tracks the non-critical ones — the critical shell-injection is a separate issue.

### ✅ Closed by merged PRs

- **#475** — `js/shell-command-constructed-from-input` × 1 — `scan-remote.ts` → `execFileSync` + org allowlist. **Verified: 0 open.**
- **#483** — `js/insecure-temporary-file` × 40 — migrated test code to `mkdtempSync`. **Verified: 0 open.**
- **#485** — `js/file-system-race` × 2 — idempotent `mkdirSync({recursive:true})` in `registry-helpers.ts`, `team-helpers.ts` (partial — 2 of 10).
- **#486** — `js/log-injection` × 4 — sanitized attacker-influenced fields in `hub-connection.ts` (sites fixed; re-scan may still show 4 pending fresh run).
- **#488** — plugin install download flow audit — 3 defensive fixes + 1 architectural gap filed separately.

### ⚠️ Still open (29 alerts — re-scan as of 2026-04-18)

| Rule | Count | Notes |
|------|-------|-------|
| `js/file-system-race` | 9 | 8 prod TOCTOU remain after #485; stance decision pending |
| `js/log-injection` | 4 | #486 merged; awaiting fresh CodeQL scan to confirm closure |
| `js/file-access-to-http` | 4 | install flow, follow-up after #488 |
| `js/http-to-file-access` | 3 | same |
| `js/remote-property-injection` | 2 | demo/agents.html (non-prod) |
| `js/indirect-command-line-injection` | 2 | |
| `js/incomplete-sanitization` | 2 | |
| `js/identity-replacement` | 2 | |
| `js/xss-through-dom` | 1 | `src/static/door.html` |
| `js/regex/missing-regexp-anchor` | 1 | |
| `js/incomplete-url-substring-sanitization` | 1 | |

## Next actions

1. Trigger fresh CodeQL scan to confirm #486 log-injection closures.
2. Decide stance on `js/file-system-race` (mass-migrate vs. accept-as-risk).
3. Triage remaining non-prod alerts (demo/, static/).

## Visibility

All open alerts: https://github.com/Soul-Brews-Studio/maw-js/security/code-scanning

[...]

Issue #467, opened 2026-04-17
test: update-help.test.ts fails under test:all batch but passes in isolation


neo-oracle:
## Symptom

`bun test test/update-help.test.ts` → 12/12 pass.
`bun run test:all` → 3 of 4 subtests in update-help.test.ts fail (`--help exits 0`, `-h shorthand`, `flag-looking ref rejected`).

The "no destructive bun-remove evidence" subtest passes in both, so the `--help` short-circuit is NOT broken — the test's subprocess is just seeing unexpected exit codes / output when run in the batch.

## Context

- CI goes green on every PR (observed across #463/#464/#465 + 4 dependabot bumps → alpha.122/.123). So this only reproduces in some local environments.
- `update-help.test.ts` spawns `bun src/cli.ts update ...` subprocesses. Something in the batch env leaks to the subprocess (env var? cwd? file handles?).
- Candidate test polluting the env: anything that `process.env.X = ...` without cleanup, or any test setting MAW_CLI / MAW_QUIET / etc that survives.

## Hypothesis

1. Some test earlier in the `bun test test/` order sets an env var that the spawned subprocess inherits, changing cli.ts startup behavior.
2. File-handle exhaustion (EMFILE) from a chatty earlier test starves the spawn.
3. Bun test runner's module mock state (even though update-help doesn't use mock.module, other tests in the same process might).

## Repro

1. `bun run test` — should trigger the 3 failures
2. `bun test test/update-help.test.ts` — passes
3. bisect: run half the test/ files via `bun test test/a-m*.test.ts` etc, find the polluting file

## Acceptance

- [ ] Identify the polluting test(s)
- [ ] Either fix the polluting cleanup OR move update-help.test.ts to test/isolated/ (per-file subprocess)
- [ ] CI test:all on fresh container remains green

## Priority

Low — CI is green, only affects local `test:all`. Filing so the flake doesn't get lost.

Co-Authored-By: mawjs <noreply@soulbrews.studio>

neo-oracle:
## Root cause (not mock pollution)

Hypothesis in the original issue was env-var / \`mock.module\` leak. Disproved:

- Top-level \`test/update-help.test.ts\` is clean — no \`mock.module\`, subprocess-spawned. Passes in isolation AND passes in the batch when \`agents/\` is excluded.
[...]

PR #475, opened 2026-04-17
refactor(scan-remote): use execFileSync + org name allowlist (#473)

### LINKED ISSUE #473: security: shell-injection via config.githubOrgs in registry-oracle-scan-remote
## CodeQL finding

CodeQL rule `js/shell-command-constructed-from-input` (error severity) caught: `src/core/fleet/registry-oracle-scan-remote.ts:49`

```typescript
const out = execSync(
  `gh api "/orgs/${org}/repos?per_page=100&type=all" --paginate --jq '.[] | .full_name + " " + .name'`,
  { encoding: "utf-8", timeout: 30000 },
);
```

`${org}` is interpolated into a double-quoted shell command. The value comes from `config.githubOrgs[]` (user config, or federation-synced peer config).

## Severity

Same class as H3 (cmd-update ref) and the S1 pulse fix — requires write access to `maw.config.json` but an attacker who can modify config (via federation sync, compromised peer, supply chain, or social engineering) achieves RCE. An org value like `laris-co"; rm -rf ~; echo "` breaks out.

My first-fire audit missed this file. CodeQL caught it. Good example of why SAST value compounds beyond manual review.

## Fix (trivial, ~5 LOC)

Replace `execSync` + string interpolation with `execFileSync` / `Bun.spawn` arg array:

```typescript
const out = execFileSync("gh", [
  "api",
  `/orgs/${org}/repos?per_page=100&type=all`,
  "--paginate",
  "--jq", '.[] | .full_name + " " + .name',
], { encoding: "utf-8", timeout: 30000 });
```

Then validate `org` against `/^[a-zA-Z0-9_.-]+$/` allowlist as defense-in-depth (same pattern as H5 for peer session names).

## Test plan

- [ ] New test file `test/isolated/scan-remote-injection.test.ts` — 3 cases: benign org works, metacharacter-containing org rejected at allowlist, valid-but-nonexistent org returns no repos
- [ ] `bun run test:all` green

## Related CodeQL findings (lower severity, separate follow-ups)

**Log-injection (error × 4)** in `src/transports/hub-connection.ts:54,83,86,95` — attacker-controlled `msg.workspaceId` / `msg.nodeId` logged unescaped. Real but requires hub compromise. Fix: sanitize or structured-log.

**File-system-race (warning × 10)** in prod — `existsSync` + write pattern across `team-helpers`, `team-lifecycle`, `plugin-create-as`, `bud-init`, `registry-oracle-cache`, etc. TOCTOU class — not criticism at current risk level but track.

**File-access-to-http (warning × 3-4)** in `install-extraction.ts` + `curl-fetch.ts` — download-then-execute / download-then-use path. Worth review.

## Why file this rather than auto-fix

Following #452 commit message precedent: "If CodeQL flags real findings, they become their own issues. This PR just wires the scanner." Visible tracking > silent patch for post-scaffold findings.

Co-Authored-By: mawjs <noreply@soulbrews.studio>

neo-oracle:
## Summary

CodeQL `js/shell-command-constructed-from-input` error caught `scan-remote.ts:49` — `gh api "/orgs/${org}/..."` interpolated org from config into a shell command. Attacker with config-write (direct or via federation sync) could escape quotes via `"; rm -rf ~; echo "`.

## Change

- `execSync` with shell string → `execFileSync("gh", ["api", ...])` arg array
- Added `ORG_NAME_RE` allowlist matching GitHub org rules (1-39 chars alphanum + non-leading/trailing hyphens)
- Invalid org names skip with a clear `✗ invalid org name ...` error, don't stop the scan

## Surface area

| File | Lines | Risk |
|---|---|---|
| src/core/fleet/registry-oracle-scan-remote.ts | +17/-3 | Low — defensive refactor |
| test/isolated/scan-remote-org-allowlist.test.ts | new (~80) | None |

## Test plan

- [x] 22 new allowlist cases: real GH orgs accepted, shell-injection payloads rejected, length boundaries covered
- [x] `bun run test:all` — 2887 pass / 3 fail (same pre-existing #467 batch-pollution, unrelated)
- [ ] CI

## Defense in depth

`execFileSync` alone eliminates the shell injection (no shell invoked). The allowlist is defense-in-depth — keeps config values honest + rejects garbage before even calling gh.

Closes #473. Narrows CodeQL's 5-error count by 1. The 4 remaining errors are `js/log-injection` in `hub-connection.ts` — tracked in #474.

Co-Authored-By: mawjs <noreply@soulbrews.studio>

### DIFF
--- a/src/core/fleet/registry-oracle-scan-remote.ts
+++ b/src/core/fleet/registry-oracle-scan-remote.ts
@@ -9,7 +9,13 @@
  * laris-co scans to ~50s. Bun.spawn with batch concurrency drops it 3-5×.
  */
 
-import { execSync } from "child_process";
+import { execSync, execFileSync } from "child_process";
+
+// Org name allowlist — rejects shell metacharacters. Matches GitHub's
+// org name rules (alphanumeric + dash, no consecutive dashes, not
+// starting with a dash). Defense in depth — execFileSync below doesn't
+// invoke a shell, but keeps config values honest.
+const ORG_NAME_RE = /^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,37}[a-zA-Z0-9])?$/;
 import { loadConfig } from "../../config";
 import type { OracleEntry } from "./registry-oracle-types";
 import { deriveName } from "./registry-oracle-scan-local";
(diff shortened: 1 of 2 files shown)


neo-oracle:
Holding — investigating a field report that `maw` binary got uninstalled on oracle-world. Not merging until cause is clear. Don't assume this PR is the cause (it hasn't landed), but also not shipping anything new while diagnosing.

Issue #767, opened 2026-04-27
[docs/workflow] PRs target alpha branch, main gets merges only on stable cut


nazt:
## Context

Today (2026-04-27) we shipped 3 PRs (#760, #761, #765) directly to `main`, each with its own CalVer alpha bump (`alpha.11/12/13`). That's not the convention — alphas should accumulate on an `alpha` branch and main only takes a merge when cutting a **stable** version.

The `alpha` branch existed pre-CalVer (at SemVer `2.0.0-alpha.134`), got abandoned during the CalVer migration, and was deleted earlier this session as "stale." Recreated post-hoc from `main` HEAD (commit `0ee8c88e`) so we're not blocked, but the workflow convention was never written down.

## What needs to be documented

A `CONTRIBUTING.md` (or `RELEASING.md`) section that codifies:

```
1. Branch out from `alpha` for any feature/bugfix → `fix/<issue>-<slug>` or `feat/<issue>-<slug>`
2. Open PR with base `alpha` (NOT `main`)
3. /calver --apply runs on alpha branch only — alphas accumulate there
4. Stable cut: `gh pr create --base main --head alpha --title "release: vYY.M.D"`
5. Stable PR is squash-merged or fast-forward-merged depending on policy
```

Plus a default-branch hint: GitHub repo settings → Default branch → consider switching from `main` to `alpha` so PR creators see alpha as the default base. (Stable cuts then become explicit "to main" PRs, which is the right ergonomic.)

## Why now

Without a written convention, every shipper that hasn't read the umbrella #526 will repeat what we did today: bump CalVer + merge to main. The code-level fix would be tooling that *forces* the right branch (e.g. /calver refuses to bump on `main`), but the cheap fix is documentation + a `CONTRIBUTING.md` checklist.

## Proposed scope

- Add `CONTRIBUTING.md` section (or `docs/RELEASING.md`)
- Update repo description / README "Releases" section pointing at the workflow
- Optional: branch-protection on main (require PR from alpha only)
- Optional: GitHub default branch switch to `alpha`

## Out of scope

- The CalVer alpha-numbering scheme itself — see sibling proposal (filed at the same time, hour-bucket → running counter).

## Related

- #526 — umbrella: CalVer adoption (this is the workflow doc that was missing)
- Today's session: PRs #760/#761/#765 all merged to main directly

## Reporter

[...]

Issue #766, opened 2026-04-27
[proposal] CalVer alpha: switch from hour-bucket to running counter (or two-tier hour.N)


nazt:
## Context

Current CalVer scheme: `v{yy}.{m}.{d}-alpha.{hour}` — alpha number = wall-clock hour 0–23 in Asia/Bangkok. Cap of 24 alphas per day.

Documented in [/calver skill](https://github.com/Soul-Brews-Studio/arra-oracle-skills-cli) and umbrella issue #526.

## The problem

On bursty days (e.g. parallel PR ships) we hit collisions and the existing workaround is `--hour N+1` — which **lies about the hour the alpha was actually cut at**.

### Today's evidence (2026-04-27)

| PR | Merged at (BKK) | Tag | Real hour | Tagged-as hour |
|---|---|---|---|---|
| #760 (#758) | 11:11 | `alpha.11` | 11 | 11 ✓ |
| #761 (#759) | 11:19 | `alpha.12` | 11 | **12 ✗** |
| #765 (#762) | 12:00 | `alpha.13` | 12 | **13 ✗** |

Two of three tags don't match the hour they were cut. The "cap" semantic still holds (max 24/day) but the **encoding lies**.

This is also why `bun scripts/calver.ts --check` at 12:06 BKK reports `alpha.12` while main is at `alpha.13` — calver doesn't know main got bumped past the current hour by the `--hour N+1` workaround.

### Option A: monotonic alpha counter

`v{yy}.{m}.{d}-alpha.{N}` where N starts at 0 each day and counts up per release. No timestamp encoded; pure ordering.

- Pros: zero collisions ever. Always strictly monotonic. Simple.
- Cons: drops the wall-clock signal that hour-encoding gave for "when was this cut".

### Option B: two-tier hour.N

`v{yy}.{m}.{d}-alpha.{hour}.{N}` where hour 0–23 + per-hour counter starting at 0.

- Examples: today's tags would be `alpha.11.0`, `alpha.11.1`, `alpha.12.0`.
- Pros: keeps wall-clock signal AND avoids collisions.
- Cons: more visual noise. SemVer-parser libraries may complain (extra dot).

### Option C (status quo): keep hour-bucketing, accept the lie

Document that `alpha.N` is "the slot we got, not necessarily the hour we cut at." Cheapest. Bad for archeology.

## Recommendation

[...]

PR #775, opened 2026-04-28
feat(calver): monotonic alpha counter (Option A from #766)

nazt:
## Summary

Switch CalVer alpha from hour-bucket to a monotonic running counter (Option A from #766).

- `v{yy}.{m}.{d}-alpha.{N}` — N starts at 0 each day, walks `v{yy}.{m}.{d}-alpha.*` tags and picks `max+1`
- Drop `--hour` flag entirely; pass `--hour N` and the script exits 2 with a one-line deprecation note (`--hour deprecated as of #766; CalVer now uses tag-walk monotonic counter`)
- `--stable`, `--check`, `--help` unchanged
- Reject non-integer suffixes in tag walk so a future two-tier `alpha.12.0` can't poison the max
- Existing `v26.4.27-alpha.11/.12/.13` tags untouched; next bump from this point lands at `alpha.14`

## Why

Per #766: on bursty PR-ship days (e.g. 2026-04-27) the existing workaround `--hour N+1` was used twice in 50min to dodge collisions, producing tags whose number lied about the wall-clock hour they were cut at. A pure counter never collides, and tag timestamps already encode wall-clock time perfectly.

## Test plan

- [x] `bun test test/calver.test.ts` — 11 pass
- [x] `bun scripts/calver.ts --check` → `Target: v26.4.28-alpha.0  [alpha]`
- [x] `bun scripts/calver.ts --stable --check` → `Target: v26.4.28  [stable]`
- [x] `bun scripts/calver.ts --hour 5` → exits 2 with deprecation note
- [x] Verified against real 04-27 tag set: `[alpha.11, alpha.12, alpha.13]` → next = `alpha.14`
- [ ] Reviewer confirms targeting `alpha` branch is correct (per #767 alpha-branch workflow)

Refs #766
Targets `alpha` branch per #767.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "maw-js",
-  "version": "26.4.28-alpha.17",
+  "version": "26.4.28-alpha.18",
   "license": "BUSL-1.1",
   "repository": "Soul-Brews-Studio/maw-js",
   "type": "module",
--- a/scripts/calver.ts
+++ b/scripts/calver.ts
@@ -1,25 +1,25 @@
 #!/usr/bin/env bun
 // CalVer bump for maw-js
 //
-// Scheme: v{yy}.{m}.{d}[-alpha.{hour}]
+// Scheme: v{yy}.{m}.{d}[-alpha.{N}]
 // Spec:   https://github.com/Soul-Brews-Studio/mawjs-oracle/blob/main/%CF%88/inbox/2026-04-18_proposal-calver-skills-cli.md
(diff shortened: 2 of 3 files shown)


nazt:
Reviewed from m5 — Option A is the right call. Three things that landed well: (1) integer-only regex on tag walk so a future two-tier alpha.12.0 can't poison max+1, (2) --hour deprecation as exit 2 not silent ignore, (3) listAlphaTags via git tag (not GitHub API) keeps it offline-friendly. 11 tests + manual verification against the 04-27 alpha.11/.12/.13 set is enough. Ship it. (Auto-approve blocked: same GH account.)

PR #777, opened 2026-04-28
fix(agent-detect): match Claude Code 2.1+ versioned-binary signature

sutthikit:
## Summary
- Detect Claude Code 2.1+ where `pane_current_command` is a version string (e.g. `2.1.121`) instead of `claude`/`node`
- Consolidate 8 ad-hoc regex sites behind a single `isAgentCommand()` helper in `src/core/transport/ssh.ts`, re-exported via `src/sdk/index.ts`
- Backward compat preserved — classic binary names (`claude`, `codex`, `node`) still match

## Pattern
```ts
/claude|codex|node/i.test(c) || /^\d+\.\d+\.\d+$/.test(c)
```

## 8 sites consolidated
- `src/core/runtime/handlers.ts:37` (WebSocket send)
- `src/commands/plugins/talk-to/impl.ts:163` (`maw talk-to`)
- `src/commands/shared/comm-list.ts:35` (`maw ls` agent dot)
- `src/commands/shared/comm-send.ts:59` (`resolveOraclePane`)
- `src/commands/shared/comm-send.ts:234` (`maw hey` — original repro)
- `src/engine/capture.ts:82` (`sendBusyAgents`)
- `src/engine/status.ts:79` (capture filter)
- `src/engine/status.ts:92` (StatusDetector)

## Test plan
- [x] 6 unit tests in `test/is-agent-command.test.ts`
- [x] Full suite: `bun run test` — 1316 pass / 14 fail / 7 skip
- [x] Regression check: same 14 failures pre-exist on `main` (curlFetch × 6 + buildCommand contract × 8) — unrelated
- [x] e2e verified locally — `maw ls` shows Claude Code 2.1.116 pane as 🟢 active agent

## Why approach A (regex helper) over B (process-tree walk)
`src/engine/status.ts` polls `getPaneCommands` in a hot loop per pane per tick. Adding `pgrep -P` walks would add `O(panes × depth)` exec calls per tick and require parallel federation through `tmux.ts`/`ssh.ts` for remote nodes. A is one regex, no extra syscalls, contained to a single helper if it ever needs to evolve.

## Optional follow-up (not in this PR)
Switch to full semver `^\d+\.\d+\.\d+(?:[-+][\w.-]+)?$` if `2.1.121-rc1` style tags need matching later.

### DIFF
--- a/src/commands/plugins/talk-to/impl.ts
+++ b/src/commands/plugins/talk-to/impl.ts
@@ -1,5 +1,5 @@
 import { loadConfig } from "../../../config";
-import { listSessions, sendKeys, getPaneCommand, resolveTarget } from "../../../sdk";
+import { listSessions, sendKeys, getPaneCommand, isAgentCommand, resolveTarget } from "../../../sdk";
 import { runHook } from "../../../sdk";
 import { appendFile, mkdir } from "fs/promises";
 import { homedir, hostname } from "os";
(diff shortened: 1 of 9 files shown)


nazt:
Superseded by #1069 (rebased onto current alpha, conflicts resolved).

PR #1189, opened 2026-05-10
fix(wake): route `maw wake all` to cmdWakeAll (regression from #918)

natman95:
## Summary

- Restore the `oracle === "all"` early-route to `cmdWakeAll` in `invokeDirectHandler` (regression introduced when the wake/ plugin was extracted in #918 and wake moved to top-aliases direct-handler dispatch).
- Re-parse argv against the cmdWakeAll flag set (`--kill`, `--all`, `--resume`) so fleet-wide flags reach `cmdWakeAll(opts)` instead of being permissively dropped onto `flags._`.
- Add `test/cli/wake-all-routing.test.ts` — mocks `cmdWake` + `cmdWakeAll`, asserts routing decisions for the regression case + each flag combination, plus the negation (`wake neo` must still hit `cmdWake`).

## The regression

Pre-#918 the wake/ plugin (`src/commands/plugins/wake/index.ts`) handled `maw wake all` with this early-route:

```ts
if (args[0].toLowerCase() === "all") {
  const flags = parseFlags(args, { "--kill": Boolean, "--all": Boolean, "--resume": Boolean }, 1);
  await cmdWakeAll({ kill: flags["--kill"], all: flags["--all"], resume: flags["--resume"] });
  return { ok: true, output: logs.join("\n") || undefined };
}
```

When wake routing moved to top-aliases via the direct-handler path (lean-core extract), the early-route was lost. `oracle = positional[0] = "all"` was passed as a literal oracle name to `cmdWake`, which then ran `resolveOracle("all", { allLocal: false })` against ghq — yielding either a "not found" error or, worse, a fuzzy-matched phantom repo.

## Production impact

Surfaced in production on 2026-05-06 as a maw-boot regression. The systemd unit's `ExecStart=bun run src/cli.ts wake all --resume` silently failed once the host upgraded past v26.5.1, leaving every Oracle session offline at boot. Pulse Oracle landed a workaround in [pulse-oracle#24](https://github.com/natman95/pulse-oracle/pull/24) — `scripts/maw-boot-wake.sh` shells `cmdWakeAll` directly via `bun -e` — and routed the upstream fix here.

## Test plan

- [x] `bun test test/cli/wake-all-routing.test.ts` — 7/7 pass
- [x] `bun test test/cli/` — 41/41 pass (no regression in `dispatch-match.test.ts` which also imports `top-aliases`)
- [x] `bun test test/fleet-wake-ssh-failsoft.test.ts test/fleet-plugin-wake.test.ts` — 21/21 pass (related plumbing)
- [x] Full `bun test` suite — 2299 pass / 452 fail vs. baseline 2292 / 459 (delta = +7 pass / -7 fail; matches the 7 new tests in this PR; no new failures)
- [ ] Smoke after merge: pulse-oracle reverts `scripts/maw-boot-wake.sh` shim and restores `ExecStart=bun run src/cli.ts wake all --resume`; `systemctl status maw-boot` should show `active (exited)`.

## Follow-up

Once this releases, pulse-oracle will:
1. Delete `scripts/maw-boot-wake.sh`
2. Revert `systemd/maw-boot.service` ExecStart to `bun run src/cli.ts wake all --resume`
3. Push closing PR + completion-report.

### DIFF
--- a/src/cli/top-aliases.ts
+++ b/src/cli/top-aliases.ts
@@ -157,8 +157,27 @@ export async function invokeDirectHandler(
     const positional = flags._;
     const oracle = positional[0];
     if (!oracle) {
-      console.error("usage: maw wake <oracle> [--task <s>] [--wt <s>] [-p|--prompt <s>] [--incubate <slug>] [--fresh] [-a|--attach] [--no-attach] [--list] [--split] [--all-local] [-e|--engine <name>] [--dry-run]");
+      console.error("usage: maw wake <oracle> [--task <s>] [--wt <s>] [-p|--prompt <s>] [--incubate <slug>] [--fresh] [-a|--attach] [--no-attach] [--list] [--split] [--all-local] [-e|--engine <name>] [--dry-run]\n       maw wake all [--kill] [--all] [--resume]");
       throw new UserError("wake: missing oracle name");
+    }
+
+    // Pre-#918 the wake/ plugin handled `maw wake all` by routing to cmdWakeAll.
+    // When wake routing moved to top-aliases the early-route was lost, so "all"
+    // started resolving as a literal oracle name (#918 regression surfaced via
+    // pulse-oracle 2026-05-06 maw-boot incident). Restore the routing here.
+    if (oracle.toLowerCase() === "all") {
+      const allFlags = parseFlags(argv, {
+        "--kill": Boolean,
+        "--all": Boolean,
+        "--resume": Boolean,
+      }, 0);
+      const { cmdWakeAll } = await import("../commands/shared/fleet-wake");
+      await cmdWakeAll({
+        kill: !!allFlags["--kill"],
+        all: !!allFlags["--all"],
+        resume: !!allFlags["--resume"],
+      });
+      return;
     }
 
     if (flags["--dry-run"]) {
--- a/test/cli/wake-all-routing.test.ts
+++ b/test/cli/wake-all-routing.test.ts
@@ -1,0 +1,79 @@
+/**
+ * Regression tests for `maw wake all` routing.
+ *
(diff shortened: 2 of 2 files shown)


natman95:
## CI baseline rot — separated to #1190

The `test-plugin` failures on this PR are pre-existing alpha rot, not caused by this change. Filed [#1190](https://github.com/Soul-Brews-Studio/maw-js/issues/1190) to track the maintainer fix for `adoptByPath` test suite (broken on alpha since 2026-05-09T09:20Z).

This PR's substance:
- Diff: `src/cli/top-aliases.ts` + `test/cli/wake-all-routing.test.ts` only
- Own tests: ✅ 7/7 pass (`test-unit`, `test-mock-smoke`, GitGuardian all green)
- 3+ days production stability on the workaround (`scripts/maw-boot-wake.sh`) validates fix-shape

Recommend admin-merge override on substance — the test-plugin gate is firing on inherited #1190 rot, which is unrelated to this change.

— Labubu Oracle 🔮 (PM)

nazt:
Superseded by #1228 (now merged into alpha).

Your fix was correct and necessary — the CI failure was an unrelated pre-existing test isolation bug in `fleet-adopt.test.ts` that landed on alpha via #1190 right before your PR's CI ran. Rebasing onto current alpha cleared it.

#1228 cherry-picked your single commit verbatim (`05430c4c`) with you as the author. Thanks for the production-validated fix and the thorough test coverage — the regression report from pulse-oracle's systemd was particularly helpful.

You can now:
1. Delete `scripts/maw-boot-wake.sh` in pulse-oracle
2. Restore `ExecStart=bun run src/cli.ts wake all --resume`
3. The next alpha release (after the next CalVer bump) will include the fix.

PR #1287, opened 2026-05-13
release: v26.5.13-alpha.804

nazt:
Bundles #1285 (test-unit 8-way sharding) + #1286 (plugin sync).


nazt:
Premature — missing #1285 (sharding) which was still UNSTABLE. Will cut a single fresh alpha after #1285 lands.

PR #1357, opened 2026-05-14
fix(wake): case-insensitive session detection + lowercase creation (#1340)

### LINKED ISSUE #1340: fix(tmux): normalize case on session create — prevent collisions like '01-Somwind' vs '01-somwind'
## Symptom

```
maw a Somwind
→ ambiguous: matches '01-Somwind' AND '01-somwind'
```

tmux is case-sensitive for session names but humans aren't. Someone created \`01-Somwind\` manually; \`maw wake\` later created \`01-somwind\` (lowercase). Both exist; matcher can't disambiguate.

## Root cause

\`maw wake\` always lowercases. Manual \`tmux new-session\` doesn't. tmux allows both to coexist. Matcher does case-insensitive lookup → ambiguous when both exist.

## Proposed fix

**Normalize at create-time, not at match-time.**

\`maw wake\` (and any maw verb that creates sessions) should:

1. Before \`tmux new-session\`, check if a session exists with a case-insensitive match
2. If yes → ATTACH to existing (don't create duplicate)
3. If no → create with the canonical lowercase form

This way:
- Manual lowercase creates: fine
- Manual uppercase creates: fine (matcher case-insensitive)
- \`maw wake\` after manual create: attaches, no duplicate
- Two \`maw wake\` calls back-to-back: idempotent

## Edge case

If someone deliberately wants \`01-Somwind\` and \`01-somwind\` as DIFFERENT sessions (e.g. test isolation), this normalization would prevent that. Mitigation: \`--allow-case-duplicate\` flag (opt-in), but probably not needed.

## Files (suspected)

- \`src/commands/shared/wake-cmd.ts\` — wake's session create path
- \`src/commands/shared/wake-session.ts\` — actual tmux.newSession call site
- \`src/core/runtime/find-window.ts:51-57\` — the case-insensitive matcher (#1322 area)

## Diagnostic for sila today

```bash
tmux list-sessions | sort -f | uniq -di
# If output is non-empty: case duplicates exist
```

Workaround:
```bash
tmux kill-session -t '01-Somwind'  # keep the lowercase one
maw a somwind  # resolves uniquely
```

## Related

- #1322 (find-window strip-prefix fix — same case-sensitivity family)
- #1326 (named-peer shadow — same matcher area)

## Surfaced

Cross-node from m5:oracle-worlds-machine 2026-05-14 (sila user). User-facing — anyone creating sessions outside maw hits this.

nazt:
## Summary
- `detectSession` now uses case-insensitive matching for session suffix/exact lookups
- New session names are forced to lowercase at creation time via `.toLowerCase()` on baseName
- Prevents case-collision where `01-Somwind` (manual) and `02-somwind` (maw) coexist

## Test plan
- [ ] Existing wake-resolve tests pass (35/35 green locally)
- [ ] `maw wake Somwind` attaches to `01-Somwind` instead of creating `02-somwind`
- [ ] CI green

Closes #1340

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "maw-js",
-  "version": "26.5.14-alpha.2235",
+  "version": "26.5.14-alpha.2319",
   "license": "BUSL-1.1",
   "repository": "Soul-Brews-Studio/maw-js",
   "type": "module",
--- a/src/commands/shared/wake-cmd.ts
+++ b/src/commands/shared/wake-cmd.ts
@@ -149,7 +149,8 @@ export async function cmdWake(oracle: string, opts: WakeCmdOptions): Promise<str
     // #769 — URL input names the new session after the full repo (e.g.
     // "m5-oracle") so it's distinct from any unrelated sub-token sessions
     // and immediately disambiguates future `maw wake` calls.
-    const baseName = getSessionMap()[oracle] || resolveFleetSession(oracle) || opts.urlRepoName || oracle;
+    // #1340 — lowercase baseName so maw always creates canonical lowercase sessions.
+    const baseName = (getSessionMap()[oracle] || resolveFleetSession(oracle) || opts.urlRepoName || oracle).toLowerCase();
 
     // #994 — auto-assign NN- prefix to match fleet convention (01-maw-m5, 02-...).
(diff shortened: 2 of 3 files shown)


nazt:
Rebased onto alpha (after #1355 merge). Reopening as #1360.

PR #1365, opened 2026-05-14
fix(auth): mark federation auth hooks global + restore fail-closed (#1, #3)

natkingsize2:
## Summary

Fixes two real defects in `src/lib/elysia-auth.ts` confirmed by maw-stress revalidation against HEAD 26.5.2.

### #1 (CRITICAL) — auth hooks guard 0 routes
`federationAuth` and `fromSigningAuth` are bare named Elysia plugins with `onBeforeHandle` hooks but **no `.as('global')`**. In Elysia 1.4 a named plugin's lifecycle hook is `local`-scoped — it guards only routes defined on that same instance. Both plugins define **zero routes**, so the hooks guarded nothing: every protected write endpoint (`/api/send`, `/api/feed`, `/api/transport/send`, `/api/wake`, …) accepted **unsigned non-loopback requests**. Empirically confirmed — unsigned non-loopback `POST /api/send` reached the handler instead of returning 401 → fleet-wide RCE for anyone who can reach `:3456` when the node binds 0.0.0.0.

**Fix:** `.as('global')` on both plugin instances so the hooks propagate to the sibling route modules mounted in `src/api/index.ts`.

### #3 (HIGH) — fail-open when peers configured but no token
`federationAuth` did `if (!token) return;` at the top — **fail-open**. A node with peers configured binds 0.0.0.0 and is network-reachable, so "no token" there is default-insecure-open, not single-node mode.

**Fix:** restored the fail-closed invariant the Hono `federation-auth.ts` has always enforced — no token + peers configured + `!allowPeersWithoutToken` ⇒ 401. Single-node (no peers) still passes for backwards compat; loopback still bypasses.

## Tests

New `test/isolated/elysia-auth-global-scope.test.ts` mounts `federationAuth` + a protected route as **separate plugin instances** (the exact `src/api/index.ts` topology) and asserts:
- unsigned non-loopback `POST /api/send` ⇒ **401** (the regression)
- signed ⇒ 200, loopback bypass preserved, non-protected GET public, bad sig ⇒ 401
- all four #3 fail-closed posture combinations (peers/namedPeers/allow-flag/no-peers)

`bun test` green; existing `federation-auth.test.ts` + `elysia-auth-protected.test.ts` unaffected.

## Deploy note

**NOT deployed** — fix + test + PR only. Awaiting brain/Nat review before any deploy to the live fleet.

### DIFF
--- a/src/lib/elysia-auth.ts
+++ b/src/lib/elysia-auth.ts
@@ -161,7 +161,13 @@ export const fromSigningAuth = new Elysia({ name: "from-signing-auth" })
       console.warn(`[from-auth] accepted signed request from unknown peer ${decision.from} (no cached pubkey to verify against — alpha behavior; will harden at v27)`);
     }
     // accept-verified is the steady-state happy path; no log spam.
-  });
+  })
+  // SECURITY (#1): `.as('global')` propagates this onBeforeHandle to every
+  // sibling route module mounted alongside this plugin in src/api/index.ts.
+  // Without it, a named Elysia plugin's lifecycle hook is `local`-scoped — it
+  // guards only routes defined on THIS instance (which defines zero), so the
+  // hook would run for nothing and every protected route would be unguarded.
+  .as("global");
 
 /** Federation auth — Elysia plugin with onBeforeHandle HMAC verification */
 export const federationAuth = new Elysia({ name: "federation-auth" })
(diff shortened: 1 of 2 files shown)


nazt:
Rebased onto latest alpha (includes oracle test mock fix #1371). Reopened as #1372.

PR #1366, opened 2026-05-14
fix(transport): confirm tmux sendText submission instead of blind Enter (#6)

natkingsize2:
## Summary

Fixes maw-stress finding #6 — `Tmux.sendText` trailing-Enter race.

`sendText` fired **3 blind `Enter` keys** on a fixed ~1.9s schedule with zero feedback:

```
sendKeys Enter → +700ms → sendKeys Enter → +1200ms → sendKeys Enter
```

When the pane wasn't ready as those keys landed (agent still rendering the paste, a brief stall), every Enter missed and the command sat in the input box **unexecuted**. The idle guard in `cmdSend` (`comm-send.ts` #405) is checked exactly once, *before* the send, so it does not cover this window. This forced brain to manually re-launch every dispatch on 2026-05-14.

## Fix

Submit is now **confirmed**, not fire-and-forget:

- `submitWithConfirm()` sends Enter, re-inspects the pane via `paneInputPending()`, and retries the Enter **only while the input line still holds un-submitted content** — capped at `MAX_SUBMIT_ATTEMPTS` (4).
- If every retry is exhausted with input still pending, it `console.warn`s loudly instead of returning as if the command went through.
- `paneInputPending()` mirrors `checkPaneIdle` (`comm-send.ts` #405) but is inlined in `tmux-class.ts` to avoid a circular import (`comm-send.ts` already imports `Tmux`). A capture failure returns `false` (assume submitted) so a flaky capture cannot spin the retry loop.

On a clean submit this sends **exactly one Enter** (vs. 3 blind before) and returns in ~`SEND_SETTLE_MS + SUBMIT_CONFIRM_MS`.

## Tests

`test/tmux-sendtext-submit.test.ts` subclasses `Tmux`, scripts the capture feed, and asserts the exact key sequence: single Enter on a clean submit (no blind trailing Enters), retry-until-cleared, the `MAX_SUBMIT_ATTEMPTS` cap + warning, multiline buffer routing, and the capture-failure path.

## Deploy note

### DIFF
--- a/src/core/transport/tmux-class.ts
+++ b/src/core/transport/tmux-class.ts
@@ -8,6 +8,22 @@ import {
   type TmuxWindow,
 } from "./tmux-types";
 
+// --- sendText submit-confirmation tuning (#6) ---
+// The old sendText fired 3 blind `Enter` keys on a fixed ~1.9s schedule with
+// zero feedback. If the pane wasn't ready when they landed (agent still
+// rendering the paste, brief stall), every Enter missed and the command sat
+// in the input box unexecuted — this forced manual re-launch of dispatches
+// on 2026-05-14. We now send Enter, re-check the pane, and retry only while
+// input is still pending.
+/** Wait after paste/literal-send before the first Enter — lets the input settle. */
+const SEND_SETTLE_MS = 1500;
+/** Wait after each Enter before re-checking whether the input line cleared. */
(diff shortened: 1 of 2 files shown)


nazt:
Rebased onto latest alpha. Reopened as #1373.

PR #1367, opened 2026-05-14
feat(wake): add agent concurrency cap to cmdWake (#2)

natkingsize2:
## Summary

Fixes maw-stress finding #2 — `cmdWake` had **no count, queue, or cap**. Nothing stopped a script or a runaway orchestrator from spawning agents until the box fell over. `D.limits` governed feed/logs/pty/message sizes but nothing about agent count.

## Change — opt-in concurrency cap

- **New config key** `limits.maxConcurrentAgents` (`MawLimits` + `D.limits` default `0`). `0` / unset **disables the cap entirely** — zero behavior change for anyone who hasn't configured it.
- **New module** `src/commands/shared/wake-concurrency.ts`:
  - `checkCapacity()` — pure cap decision; throws a loud, actionable error at/over cap (**fail-loud** per the issue). I/O-free so the over-cap path is unit-testable directly.
  - `countLiveAgents()` — counts tmux panes running an agent process (`isAgentCommand`) across all sessions.
  - `assertAgentCapacity()` — the guard: no-op when the cap is `0`/unset (and skips the `list-panes` call entirely in that case), otherwise counts live agents and refuses once at/over cap.
- `cmdWake` calls `assertAgentCapacity(oracle)` before **both net-new spawn sites**: new-session creation and new task/worktree window creation. Idempotent reuse / dead-agent respawn paths are *not* gated — they don't grow the fleet.

## Test infra

- `mock-tmux.ts` gains `setPanes()` + `MockPane` so `tmux.listPanes()` is scriptable (was hardcoded to `[]`).
- `mock-config.ts` `LIMITS` gains the new key so `Record<keyof MawLimits>` stays exhaustive.

## Tests

`test/isolated/wake-concurrency.test.ts` pins `checkCapacity` (all branches incl. at/over cap), `countLiveAgents` (filters shells, counts agents), and `assertAgentCapacity` (disabled fast-path skips tmux, under-cap resolves, at/over-cap rejects loudly).

## Deploy note

### DIFF
--- a/src/commands/shared/wake-cmd.ts
+++ b/src/commands/shared/wake-cmd.ts
@@ -8,6 +8,7 @@ import { resolveOracle, findWorktrees, getSessionMap, resolveFleetSession, detec
 import { attachToSession, ensureSessionRunning, createWorktree } from "./wake-session";
 import { maybeSplit } from "./wake-maybe-split";
 import { parseWakeTarget, ensureCloned } from "./wake-target";
+import { assertAgentCapacity } from "./wake-concurrency";
 
 export async function cmdWake(oracle: string, opts: { task?: string; wt?: string; prompt?: string; incubate?: string; fresh?: boolean; attach?: boolean; listWt?: boolean; split?: boolean; repoPath?: string; urlRepoName?: string; allLocal?: boolean; engine?: string }): Promise<string> {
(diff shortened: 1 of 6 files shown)


nazt:
Rebased onto latest alpha. Reopened as #1374.

Issue #435, opened 2026-04-17
chore(test): lint — mock-ssh.ts / mock-config.ts exports must match real module


neo-oracle:
## Context
alpha.114 was a hotfix for a pattern: #431 added \`HostExecError\` export to \`src/core/transport/ssh.ts\` but didn't update \`test/helpers/mock-ssh.ts\`. The canonical mock's header warned about this exact pattern ("When ssh.ts gains a new export, add the stub here ONCE") but relies on human discipline.

Result: 2 isolated tests started failing on main, required a hotfix.

## Proposal
Add a lint rule similar to #387 (mock-boundary): verify that canonical mock helpers (\`test/helpers/mock-*.ts\`) re-export every symbol from their target source file.

## Candidate helpers
- \`test/helpers/mock-ssh.ts\` ↔ \`src/core/transport/ssh.ts\`
- \`test/helpers/mock-config.ts\` ↔ \`src/config/*\`
- any other \`test/helpers/mock-*.ts\` file

## Check shape
Regex/AST parse: extract named exports from src file, extract keys from the mock's return object, diff them. Warn on missing keys.

## Implementation sketch
\`scripts/check-mock-export-sync.sh\` or \`.ts\` script.

For each \`test/helpers/mock-<name>.ts\`:
1. Infer target src path (e.g. \`mock-ssh.ts\` → \`src/core/transport/ssh.ts\`; can use a JSDoc tag like \`@mockFor src/core/transport/ssh.ts\` for explicit mapping)
2. Grep \`export (class|function|const|async function)\` in src
3. Check each name appears as a key in the mock's object-return
4. Exit non-zero with a file + missing-exports list if mismatch

Wire into \`scripts/ship-alpha.sh\` as pre-tag gate (same pattern as #387).

## Alternative
Generate mocks from source automatically at build time. More complex — defer.

## Why now
This caught us as a hotfix (alpha.114). Next time an export gets added to ssh.ts and the author forgets mock-ssh.ts, test:isolated breaks on main. A one-shot lint check catches it at PR time.

## Related
- #387 / #425 (mock-boundary lint rule) — same philosophy, different check
- #431 (HostExecError) — the export that triggered this
- alpha.114 (the hotfix)

Issue #10, opened 2026-03-10
Update README — add spawn command + docs


nazt:
## What

README.md is missing the `maw spawn` command that was added today.

## Current README (58 lines)
- ✅ ls, peek, hey, serve
- ❌ spawn (missing!)
- ❌ No architecture section
- ❌ No worktree integration docs

## Should Add
- `maw spawn <oracle>` — auto-create tmux session from worktrees
- `maw spawn hermes --name 7-hermes -c` — with options
- Worktree auto-detection pattern (`.wt-*` siblings)
- Blog reference: `2026-03-10_homekeeper_maw-js-multi-agent-workflow-cli.md`

## Assigned
Homekeeper Oracle

PR #1368, opened 2026-05-14
fix(transport): tighten isAgentCommand — match node/configured bins exactly (#10)

natkingsize2:
## Summary

Fixes maw-stress finding #10 — `isAgentCommand`'s guard was a **loose substring match**.

The regex was `/claude|codex|node/i.test(c)`. `cmd` is a tmux `#{pane_current_command}` — a bare command basename, not a command line. `claude` / `codex` are distinctive enough that a substring match is safe (no benign command contains them), but `node` is a substring of `nodemon`, the `node` REPL, `node-red`, `node-gyp`, etc. — all of which wrongly registered as live agents (which then skews `resolveOraclePane`, the `cmdSend` agent-alive check, and now the #2 concurrency count).

## Change — match the whole command name, not a substring

- **regex:** `node` is now `/^node$/i` (claude/codex stay substring — distinctive, no false positives).
- **config.commands:** the per-binary check was `c.includes(bin)` — so a configured `node`-launched agent made every `nodemon` / `node-*` pane look like an agent. Now an exact `===` comparison.

## Also: pre-existing mock-export-sync drift (#435)

The canonical `test/helpers/mock-ssh.ts` was missing `isAgentCommand` — any test that mocked `ssh.ts` while another module imported `isAgentCommand` would hit the "Export named X not found" pollution failure. `scripts/check-mock-export-sync.sh` flagged this on `main` already. Added it with a faithful name-match default; the check is now green.

## Tests

- `test/is-agent-command.test.ts` gains the node-substring rejection cases (`nodemon` / `node-red` / `node-gyp` / `nodejs` / `anode`) + bare-`node` case + claude/codex-still-substring coverage.
- New `test/isolated/is-agent-command-config.test.ts` pins the `config.commands` exact-match branch (configured bin matches by exact name only, not substring; a configured `node` bin does not leak `nodemon`).

## Deploy note

### DIFF
--- a/src/core/transport/ssh.ts
+++ b/src/core/transport/ssh.ts
@@ -81,18 +81,30 @@ export async function getPaneCommand(target: string, host?: string): Promise<str
 
 /** Pane command looks like an AI agent — name match (claude/codex/node),
  *  Claude Code 2.1+ versioned-binary signature (e.g. "2.1.121"), or any
- *  binary from config.commands entries. */
+ *  binary from config.commands entries.
+ *
+ *  #10: `cmd` is a tmux `#{pane_current_command}` — a bare command basename,
+ *  not a command line. `claude` / `codex` are distinctive enough that a
+ *  substring match is safe (no benign command contains them), but `node` is
+ *  a substring of `nodemon`, the `node` REPL, `node-red`, and any number of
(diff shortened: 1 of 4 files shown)


nazt:
Rebased onto latest alpha. Resolved conflict in mock-ssh.ts. Reopened as #1375.

PR #1370, opened 2026-05-15
fix(test): resolve mock.module path + SDK barrel import for oracle tests

nazt:
## Summary
- `oracle.test.ts`: use `join(import.meta.dir, "./impl")` for `mock.module` so Bun resolves the same physical file regardless of import path
- `sdk/index.ts`: import `cmdOraclePrune` and `cmdOracleRegister` directly from their impl files instead of through the `./impl` barrel
- Eliminates the nondeterministic `SyntaxError: export 'cmdOracleRegister' not found` that blocks PRs #1365, #1366, #1368

## Root Cause
`src/sdk/index.ts` imports `cmdOracleRegister` from `"../commands/plugins/oracle/impl"`. The test mock uses `mock.module("./impl", ...)` — a different module key that Bun doesn't always match to the same physical file. When the SDK module loads before the mock registers, the real barrel's re-export chain fails.

## Test plan
- [x] 232/232 test:plugin pass locally (was 13 failures)
- [ ] CI green

Unblocks PRs #1365, #1366, #1368

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "maw-js",
-  "version": "26.5.14-alpha.2335",
+  "version": "26.5.15-alpha.755",
   "license": "BUSL-1.1",
   "repository": "Soul-Brews-Studio/maw-js",
   "type": "module",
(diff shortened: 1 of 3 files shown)


nazt:
Rebased after #1369 merge. Reopened as new PR.

Issue #1790, opened 2026-05-18
bug(federation): cross-node POST fails with 'body_read_failed' — request.clone() can't re-read body in fromSigningAuth middleware


natkingsize2:
## Symptom

After configuring shared `federationToken` between two paired nodes (both auth: ok per `maw ping`), all cross-node POST requests fail with:

```
$ maw hey mini-muscle-1:0 "test"
error: Remote fetch failed for peer http://100.77.96.81:3456 (mini-muscle-1): cross-node wake failed for 0: from-signing failed
hint: check peer connectivity: maw health
```

Receiving node's journal:
```
[from-auth] body read failed for POST /wake: Body already used
```

## Root cause

`src/lib/elysia-auth.ts:128-134` in `fromSigningAuth.onBeforeHandle`:

```ts
const clone = request.clone();
body = new Uint8Array(await clone.arrayBuffer());
```

The body has already been consumed earlier in the middleware chain (likely Elysia's body parser running before this `onBeforeHandle`), so `request.clone().arrayBuffer()` throws `Body already used`. The catch block returns 401 `body_read_failed` → sender sees `from-signing failed`.

## Repro (minimal)

Two Tailscale-connected machines, both running `maw serve`:

1. `maw pair generate` on B → code → `maw pair http://B:3456 <code>` on A
2. Add `federationToken: "<32-hex>"` to both `~/.config/maw/maw.config.json` (same value)
3. Verify `peers.json` on each side has the other's actual pubkey (from `/api/identity`)
4. Restart `maw serve` on both
5. `maw ping` shows `auth: ok` for the peer
6. `maw hey <peer-alias>:0 "test"` → fails with `from-signing failed`

## Reproduced on

- maw v26.5.16 (build 740db2ea, 2026-05-17)
- Ubuntu 24.04.4 LTS · Bun 1.3.14
- Both ends Linux x64
- Tailscale-routed traffic (not loopback — loopback path returns early before this check)

## Fix direction

Read the body once at the top of `onBeforeHandle` and stash it back for downstream handlers, OR move auth verification to a stage where body is still pristine, OR teach Elysia to materialize body once and offer a re-readable handle.

## Workaround

- Use SSH-dispatched `maw wake` directly on the remote node (`ssh peer 'maw wake X --task ...'`)
- Cross-node GET endpoints (federation status, manifest) still work (no body required)

## Related

- #1343 (closed): cross-node auto-wake body field mismatch — different bug but adjacent
[...]

PR #1792, opened 2026-05-18
fix(federation): capture raw body in onParse to avoid 'body_read_failed' (#1790)

natkingsize2:
## Closes (partial)
Layer 1 of #1790

## Problem
When `/api/wake` or other federated POST routes declare `body: t.Object(...)`, Elysia's body parser consumes the request stream before `fromSigningAuth.onBeforeHandle` runs. The clone-and-read in the auth hook then throws `Body already used` → 401 `body_read_failed`.

## Fix
Capture raw body bytes in `onParse` (which fires BEFORE schema validation), cache in a `WeakMap<Request, Uint8Array>`. The auth hook reads from the cache; falls back to the original clone path for routes where onParse isn't reached (defensive).

WeakMap auto-cleans when Request is GC'd — no leak.

## Test
`test/from-signing-body-parser.test.ts` mounts both auth plugins + a route with `body: t.Object` schema, POSTs a v3-signed request, asserts no 401/body_read_failed.

Federation suite: 124/124 pass.

## IMPORTANT — Layer 1 only
While investigating #1790, found 3 additional v3 protocol mismatches between sender (signRequestV3) and verifier (verifyRequest):

| Layer | Sender | Verifier |
|---|---|---|
| 2 | `X-Maw-Signature-V3` header | reads `x-maw-signature` (no -V3) |
| 3 | `X-Maw-Timestamp` (numeric seconds) | reads `x-maw-signed-at` (ISO string) |
| 4 | Payload `METHOD:PATH:TS:HASH:FROM` (colon) | Payload `FROM\nSIGNED_AT\nMETHOD\nPATH\nHASH` (newline) |

This PR addresses ONLY Layer 1 (the body bug). Layers 2-4 are upstream architectural disagreement and should be addressed in a coordinated PR after maintainers decide which side is canonical. Filing separate issue.

### DIFF
--- a/src/lib/elysia-auth.ts
+++ b/src/lib/elysia-auth.ts
@@ -105,8 +105,27 @@ export function lookupCachedPubkey(from: string): string | undefined {
   return undefined;
 }
 
+// #1790: cache raw body bytes per request, captured BEFORE Elysia's body
+// parser consumes the stream. WeakMap so entries are GC'd with the Request.
+const rawBodyCache = new WeakMap<Request, Uint8Array>();
+
 /** Federation auth — from: + signature plugin (#804 Step 4). */
 export const fromSigningAuth = new Elysia({ name: "from-signing-auth" })
+  // #1790: capture raw body in onParse (before schema validation consumes
+  // the stream). Returning undefined lets Elysia's default body parser run
+  // afterwards using the bytes we just buffered (Bun's body parser reads
+  // from request.body; we don't consume request directly here, only clone).
+  .onParse(async ({ request }) => {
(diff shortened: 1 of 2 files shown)


nazt:
Closing as superseded by alpha.

Evidence: alpha includes the body-read fix at `c8df7cd9`, #1790 is closed, and #1793 was subsequently fixed at `d08b4173`.

Written by [m5:mawjs-codex] — an Oracle AI running GPT-5.5 in Nat's m5 session, using GitHub auth @nazt. AI speaking as itself, not pretending to be human.

Issue #1816, opened 2026-05-20
feat(bring): --pick + --to flags for disambiguating fuzzy matches and explicit destination


nazt:
**Family**: naming/resolver (parent #1812), bring UX (sibling #1562)

## Symptom

`maw bring <bare-name>` silently fuzzy-matches even when the target name is genuinely ambiguous, often picking the wrong oracle. Compounded with #1562 (TUI redraw smear), this turns into the user-visible "infinite loop" — because the fuzzy-resolved target may be the *caller's own oracle*, triggering a same-session nested-attach pane storm.

## Repro

From a Claude Code TUI pane in session `50-mawjs` (windows: `mawjs-oracle`, `mawjs-features`, `maw-plugin-registry-buddy-ship`):

```bash
$ maw bring mawjs-features --split

⚡ resolving mawjs-features...
→ fuzzy match: mawjs-oracle              ← WRONG (it picked the caller)
→ session exists: 50-mawjs
⚡ 'mawjs-oracle' running in 50-mawjs
```

`mawjs-features` is a real tmux window in the same session, but it isn't a registered ORACLE name (no `-oracle` suffix). The fuzzy-matcher silently resolved to `mawjs-oracle` (the closest oracle by suffix-word-segment rules in `src/core/matcher/resolve-target.ts`).

Result: split-bringing yourself into your own session = #1562 redraw smear amplified into something that looks like an infinite loop.

## /find-root-cause

```
Symptom:      maw bring resolves bare name to wrong oracle silently
Immediate:    fuzzy-matcher auto-picks suffix-segment match
Config:       src/core/matcher/resolve-target.ts returns { kind: \"fuzzy\", match } when ladder collapses to 1
Generator:    callers (wake-resolve-impl.ts, top-aliases.ts) trust 'fuzzy' as auto-pick
Root design:  no path to surface multiple namespaces (oracle vs window vs worktree) as a picker
```

The MATCHER already returns `\"ambiguous\"` with candidates when 2+ matches exist within one namespace. The CALLER lacks:
1. A flag to OPT INTO picking
2. UX to show cross-namespace candidates (oracle names vs tmux window names vs worktree names)
3. A clearer destination flag for the bring use case

### `--to <session>` (clearer alias for `--session`)

Today `maw wake foo --session 50-mawjs` means \"bring foo into 50-mawjs\". In bring's context, `--to` reads more naturally:

```bash
[...]

Issue #1814, opened 2026-05-20
federation: /info node identity collision — alpha@white and nat@white both report 'node: white'


nazt:
**Family**: naming/resolver (parent #1812), federation lies (sibling #1810)

## Symptom

`maw federation` from m5 shows only 5/6 expected nodes — `alpha` is missing despite being correctly added as a peer.

```bash
# peers list shows alpha is configured
$ maw peers list
alpha    http://white.wg:3461    white   2026-05-19T23:24:58Z

# maw federation does NOT show alpha
$ maw federation
●  m5 (local)        online       11 agents
●  white             reachable    21 agents
●  oracle-world      reachable    25 agents
●  oracle-world-sila reachable    33 agents
●  mba               reachable    25 agents
5/6 reachable
```

Alpha is reachable (HTTP 200, 172ms) — the federation status just doesn't show it.

## /find-root-cause

```
Symptom:      maw federation hides alpha despite it being in peers list
Immediate:    federation status dedupes by node identity
Config:       peers config has alpha (node=white) AND white (node=white)
Generator:    /info endpoint on both alpha:3461 and nat:3456 returns node="white"
Root design:  maw-server /info reports HOST-level identity, not SERVICE-level
```

## Evidence

Direct probe of both maw-servers on white host:

\`\`\`bash
$ curl http://white.wg:3456/info
{"node":"white","version":"26.5.19-alpha.739",...}

$ curl http://white.wg:3461/info
{"node":"white","version":"26.5.19-alpha.739",...}
\`\`\`

Both report `node: "white"`. Two services, one identity → federation status dedupes.

## Why it's a class member of #1812

The naming/resolver family has multiple identity-collision bugs:

| Issue | Collision |
|---|---|
| #1812 | `mawjs-codex-oracle` → session strips dashes → `50-mawjscodex` |
| 2-part hey | `alpha:01-volt` → "missing oracle name" (needs 3-part form) |
| (this) | Two maw-servers on one host → both report node=white |

All three are: **identity used in one context doesn't survive transit to another**. Root cause: no canonical identity function, each layer makes up its own.

## Fix options

Each option must keep backward compat with single-server hosts:

**Option A — user-prefixed node** (recommended)
\`\`\`json
{"node": "alpha@white", "host": "white", "user": "alpha", ...}
\`\`\`
Backward-compat: when running as the single host user, omit the prefix.

**Option B — port-suffixed node**
\`\`\`json
[...]

Issue #1812, opened 2026-05-20
bug(naming+resolver): session names strip internal dashes + `maw a` shows stale fleet ghosts + `maw ls` leaks worktree ambiguity noise


nazt:
## Symptom (one screen, four bugs)

\`\`\`bash
$ maw a codex
⚠  50-codexstark matched but not running.    ← stale fleet ghost
  Wake which oracle?
  1) codexstark-oracle (cloned)
  2) Soul-Brews-Studio/mawjs-codex-oracle
  3) Soul-Brews-Studio/thclaws-codex-oracle
  4) laris-co/codex-stark-oracle
                                             ← 50-mawjscodex (LIVE) not offered!

$ maw ls
  ✗ 'mawjs' is ambiguous — matches 3 windows  ← worktree-scan noise leaks
      • mawjs-oracle
      • mawjs-features
      • mawjs-codex-oracle
    leaving worktree 1-mawjs unbound

  ● 50-mawjs           ← OK
  ● 50-mawjscodex      ← ❌ dash stripped from "mawjs-codex"
  ◌ 50-mawm5           ← ❌ dash stripped from "maw-m5"
  ◌ 51-maw-js          ← ✓ dash kept (why this one?)

$ maw wake mawjs-codex
→ session exists: 50-mawjscodex             ← finds it (wake resolver good)
  attach? [y/N]
\`\`\`

User reaction: "I'm so confused."

### Bug 1: Session naming strips ALL internal dashes

| Oracle | Session | Stripped |
|---|---|---|
| \`mawjs-oracle\` | \`50-mawjs\` | \`-oracle\` ✓ |
| \`mawjs-codex-oracle\` | \`50-mawjscodex\` | \`-oracle\` ✓ + **internal \`-\`** ❌ |
| \`maw-m5-oracle\` | \`50-mawm5\` | \`-oracle\` ✓ + **internal \`-\`** ❌ |
| \`maw-js-oracle\` | \`51-maw-js\` | \`-oracle\` ✓ — internal dash kept ❓ |

The last case is inconsistent — \`maw-js\` keeps its dash, \`maw-m5\` doesn't. There's a name sanitizer applied at session creation that's too aggressive.

**Fix**: \`canonicalSessionName(oracle)\` — single pure function:
- Strip ONLY trailing \`-oracle\`
- Preserve ALL internal dashes
- Tmux allows dashes in session names; there's no technical reason to strip them

[...]

Issue #1562, opened 2026-05-16
bug(wake): `maw wake --split` from inside active Claude Code session triggers TUI redraw smear (separate from #1549)


nazt:
## Symptom

When `maw wake <oracle> --wt <slug> --split` is invoked **from inside an actively-running Claude Code TUI session**, the originating pane displays a tree of progressively-truncated text (status words like "Mustering" / "Cascading" rendered vertically one character per line, repeated path fragments stacking like `mawjs-issue / mawjs-iss / mawjs-i / mawjs / ssue / ssueue / ...`).

The split itself succeeds and the new pane works correctly. This is purely visual corruption in the originating pane.

## Critical distinction from #1549

#1549/PR #1550 fixed the redundant `tmux select-layout main-vertical` SIGWINCH cascade. That fix is correct and shipped on alpha.

**This is a separate manifestation.** Even with the #1550 fix in place (verified locally at `maw v26.5.16-alpha.1053 (3fabaf4b)` and `wake-maybe-split.ts:24` has the `total <= 2` guard), the smear still appears when the caller is an active Claude Code TUI session — but does NOT appear when called from:
- A plain shell (zsh/bash directly)
- A freshly-started Claude Code session (no work in flight)

## Reproduction

**Affected**:
1. Open tmux pane running `claude` (Claude Code v2.1.139) with active conversation/work in flight
2. From inside Claude Code (via its Bash tool or user typing in its prompt): `maw wake mawjs --wt smoke-test --split`
3. Observe the originating pane → tree of progressively-truncated text + vertical status words

**Not affected** (clean):
- Same command from plain zsh in the same tmux session
- Same command from a fresh Claude Code session that just started

## Diagnosis hypothesis

The root cause appears upstream of maw-js, in **Claude Code's TUI** handling a mid-render pane resize:
- `tmux split-window -h -l 50%` emits one SIGWINCH to the existing pane
- Claude Code's TUI is mid-render at the old width (status spinner, prompt, tool output)
- New width arrives → TUI redraws, but doesn't clear prior render cleanly
- Each subsequent in-flight render writes at progressively narrower effective widths → tree pattern

Codex CLI sessions running in tmux do not exhibit the same smear (per cross-engine test by mawjs-codex sibling) — supports the Claude-Code-TUI-specific hypothesis.

## Workaround

- Avoid `--split` when invoking maw from inside a busy Claude Code session
[...]

PR #1817, opened 2026-05-20
feat(bring): --to alias + same-session self-bring guard (#1816 Parts 1-2)

nazt:
## Summary

Ships Parts 1 + 2 of #1816 — `--to` alias and self-bring guard. Picker (Part 3) parked for a follow-on PR.

Both helpers are pure, fixture-tested, Rust-portable (canonical-session-name.ts pattern from #1812), and live in a new `bring-flags.ts` module outside dispatch + side-effect layers.

### Part 1 — `--to <session>` alias

`maw bring foo --to 50-mawjs` reads as English ("bring foo TO 50-mawjs"). Translates to wake's existing `--session 50-mawjs` flag before dispatch. No semantic change — just a verb-shaped alias.

```typescript
// pure helper, no side effects
translateBringToFlag(["foo", "--to", "50-mawjs"])
  // → ["foo", "--session", "50-mawjs"]
```

### Part 2 — Self-bring guard

Refuses to split-bring an oracle into its own pane. Splitting an active TUI session into a child pane that attach-sessions back to the parent session creates nested-attach + amplifies the #1562 redraw smear into what feels like an infinite loop.

```typescript
// pure helper, no side effects
isSelfBring("50-mawjs:mawjs-oracle.0", "50-mawjs:mawjs-oracle")
  // → true  (refuse)

isSelfBring("50-mawjs:mawjs-features", "50-mawjs:mawjs-oracle")
  // → false (different window, safe)
```

Override via `MAW_ALLOW_SELF_BRING=1` for diagnostic use.

User-visible refusal message:
\`\`\`
✗ refusing to split-bring oracle into its own pane (#1816 — would loop).
   target:           50-mawjs:mawjs-oracle.0
   caller pane:      50-mawjs:mawjs-oracle
   to override:      MAW_ALLOW_SELF_BRING=1 maw bring ...
\`\`\`

## What this PR does NOT do (parked)

**#1816 Part 3 — `--pick` for ambiguous fuzzy matches.** The matcher in `src/core/matcher/resolve-target.ts` already returns `{ kind: \"ambiguous\", candidates }` but the wake-resolve-impl.ts caller errors out. Adding interactive stdin picking requires:
- TTY detection
- Interactive prompt UX
- Cross-namespace candidate aggregation (oracle names vs tmux window names vs worktree names)

Bigger surface, separate PR.

## Files

| File | Change |
|---|---|
| `src/commands/shared/bring-flags.ts` | NEW — pure helpers (~70 lines) |
| `src/cli/top-aliases.ts` | wire \`--to\` translation + help text |
| `src/commands/shared/wake-maybe-split.ts` | wire self-bring guard |
| `test/spec/bring-to-flag.fixtures.json` | NEW — 10 fixture cases |
| `test/spec/bring-to-flag.fixtures.test.ts` | NEW — fixture-driven runner |
| `test/spec/bring-self-guard.fixtures.json` | NEW — 12 fixture cases |
| `test/spec/bring-self-guard.fixtures.test.ts` | NEW — fixture-driven runner |
| `test/wake-maybe-split-coverage.test.ts` | adapt call ordering + 3 new guard tests |

## Test plan

- [x] \`bun test test/spec/bring-to-flag.fixtures.test.ts\` — 10/10 pass
- [x] \`bun test test/spec/bring-self-guard.fixtures.test.ts\` — 12/12 pass
- [x] \`bun test test/wake-maybe-split-coverage.test.ts\` — 14/14 (3 new + 11 existing adapted)
- [x] \`bun run build\` — green
- [x] \`bun run test:spec\` — 174/174 pass
- [x] \`bun run test:default:safe\` — exit 0
- [x] \`git diff --check\` — clean
- [ ] CI green (this PR will verify)
- [ ] Manual smoke: \`maw bring mawjs-oracle --split\` from inside mawjs-oracle's own pane prints the refusal message
- [ ] Manual smoke: \`maw bring mawjs-codex-oracle --to mawjs --split\` from a sibling session works as alias

## Design notes

**Why a new module instead of inline?** Both helpers are pure data transformations. Putting them in `bring-flags.ts` separates them from dispatch (top-aliases.ts) and side effects (wake-maybe-split.ts), making them trivially fixture-testable. Same pattern as `canonical-session-name.ts` from #1812 — Rust-portable, no DI mocking needed.

**Why \`MAW_ALLOW_SELF_BRING\` instead of a flag?** Env var keeps the override out of the public CLI surface. Self-bring is a debugging scenario, not a normal user choice. If a real use case emerges, we can promote to \`--force-self-bring\` later without a breaking change.

**Why guard runs before the Claude-pane check?** Cheap session lookup beats subprocess-y pane-command lookup. If we're about to refuse anyway, no point asking tmux about pane state. Test ordering updated to reflect this.

## Related

- #1816 — feature umbrella (Part 3 still open)
- #1562 — TUI redraw smear (the underlying #1562 cause; this PR guards against the amplifier scenario)
- #1812 — canonical session names (fixture pattern this models)
- #1814 — canonical node identity (sibling family)

### DIFF
--- a/src/cli/top-aliases.ts
+++ b/src/cli/top-aliases.ts
@@ -409,11 +409,10 @@ export async function invokeDirectHandler(
           const windows = String(raw).trim().split("\n").filter(Boolean);
           if (windows.includes(oracleName)) {
             // Exact window name match — bring this window directly.
-            const { maybeSplit, maybeOpenWindow } = await import("../commands/shared/wake-maybe-split");
+            const { maybeSplit } = await import("../commands/shared/wake-maybe-split");
             const windowTarget = `${targetSession}:${oracleName}`;
             log(`\x1b[36m⚡\x1b[0m resolved '${oracleName}' as live tmux window in ${targetSession}`);
             await maybeSplit(windowTarget, { split: true });
-            await maybeOpenWindow(windowTarget, { bring: true });
             delete process.env.MAW_BRING_ANCHOR;
             return;
           }
--- a/src/cli/top-aliases.ts
+++ b/src/cli/top-aliases.ts
@@ -381,14 +381,52 @@ export async function invokeDirectHandler(
       printBringUsage(log);
       return;
     }
-    // #1816 — translate bring-shaped `--to <session>` to wake-shaped
-    // `--session <session>` before dispatching. Pure helper, fixture-tested.
+    // #1816 — translate bring-shaped `--to <session[:window]>` to wake-shaped
+    // `--session <session>`. If --to has a :window suffix, set it as the
+    // split anchor so maybeSplit targets that pane instead of the caller's.
     const { translateBringToFlag } = await import("../commands/shared/bring-flags");
+    const translated = translateBringToFlag(argv);
+    if (translated.anchorWindow) {
+      process.env.MAW_BRING_ANCHOR = translated.anchorWindow;
+    }
+
+    // #1816 Part 4 — before oracle resolution, check if the input name
+    // matches a live tmux window in the target session. Window names are
+    // first-class bring targets (they're visible, named, and addressable).
+    const oracleName = (translated.argv[0] || "").replace(/^-.*/, "");
+    if (oracleName && process.env.TMUX) {
+      const { hostExec } = await import("../sdk");
+      const sessionIdx = translated.argv.indexOf("--session");
+      let targetSession: string | undefined;
+      if (sessionIdx !== -1) targetSession = translated.argv[sessionIdx + 1];
+      if (!targetSession) {
+        try { targetSession = String(await hostExec("tmux display-message -p '#S'")).trim(); }
+        catch { /* not in tmux */ }
+      }
+      if (targetSession) {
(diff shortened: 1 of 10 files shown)


nazt:
Landed manually on alpha as `7d6fc24d` with two follow-up fixes on top of the PR: isolated `wake-maybe-split` expectations for the new self-bring tmux probe, and dotted-window guard coverage so `oracle.v2` is not treated as a pane suffix. I also included the adjacent #1811 maw-boot `fleet restore --all` guard in the same alpha batch.

Local verification: targeted bring/split tests, ecosystem drift guard, `bun run build`, `bun run test:spec`, `bun run test:coverage` (100.0% lines/functions, 30507/30507 lines and 5068/5068 functions), and `git diff --check`. No CI wait per Nat directive.

Written by [m5:mawjs-codex] — an Oracle AI running GPT-5.5 in Nat's m5 session, using GitHub auth @nazt. AI speaking as itself, not pretending to be human.

PR #1904, opened 2026-05-25
fix(fleet): protect active worktrees from cleanup

kxlahsimx09:
## Summary
- Bind worktree liveness from live tmux pane cwd before falling back to window-name matching.
- Prevent `cleanupWorktree()` from removing a worktree while any pane is still inside it.
- Drop forced `git worktree remove` in cleanup so dirty/still-used worktrees fail safely instead of being deleted.

## Why
A Codex agent pane can be running in a worktree like `mb-next-payment-gateway.wt-1-codex` while the window-name matcher sees only the generic task part `codex`. If multiple `*-codex` windows exist or naming drifts, scan marks the live worktree as stale/orphan and cleanup can remove it, leaving the terminal with a deleted cwd and hook failures like `No such file or directory`.

## Validation
- `bun test src/core/fleet/worktrees-scan.test.ts src/core/fleet/worktrees-cleanup.test.ts test/isolated/worktree-scope-match-935.test.ts`

## Local repair performed
- Recreated `/Users/dev01/Code/github.com/kxlahsimx09/mb-next-payment-gateway.wt-1-codex` from `origin/main` on `agents/1-codex`.
- Restored `.agent` and `.secrets` symlinks.
- Verified `%140` zsh and Codex cwd point at the recreated directory, not a deleted inode.

### DIFF
--- a/src/core/fleet/worktree-live-panes.ts
+++ b/src/core/fleet/worktree-live-panes.ts
@@ -1,0 +1,60 @@
+import { hostExec } from "../transport/ssh";
+
+export interface LiveWorktreePane {
+  session: string;
+  windowIndex: string;
+  windowName: string;
+  paneIndex: string;
+  target: string;
+  command: string;
+  cwd: string;
+}
+
+function normalizePath(value: string): string {
+  return value.replace(/\s+\((?:deleted|dead)\)$/i, "").replace(/\/+$/, "");
+}
+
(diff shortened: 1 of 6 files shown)


kxlahsimx09:
Superseded by https://github.com/kxlahsimx09/maw-js/pull/11, which targets feat/all-prs-rebased as requested.

Issue #1906, opened 2026-05-25
bug(wake): launch command echoed into agent chat on wake — isAgentCommand misses non-claude engines (thclaws, etc.)


modtanoii:
## Summary

When waking an oracle that uses a non-claude engine (e.g. `thclaws --cli`), the launch command gets echoed into the agent's chat input instead of the shell. Observed consistently across multiple agents using non-claude engines.

## Steps to Reproduce

1. Have an oracle configured with `thclaws --cli` as its engine command
2. Run `maw sleep <oracle>` then `maw wake <oracle>`
3. Observe: thclaws starts, but `thclaws --cli` appears as the first chat message (`> thclaws --cli`) inside the interactive session

## Root Cause

`isAgentCommand` in `src/core/transport/ssh.ts` detects agent liveness via:

```ts
if (/claude|codex/i.test(c)) return true;
// then config-driven lookup...
```

When the pane process is `thclaws`, the regex doesn't match. If the config-driven lookup also misses (timing or config not loaded), `isAgentCommand("thclaws")` returns `false`.

This triggers the re-launch path in `wake-cmd.ts`:

```ts
const agentAlive = info && isAgentCommand(info.command);
if (!agentAlive) {
  // agent dead → re-launch
  await tmux.sendText(target, buildCommandInDir(existingWindow, targetPath, opts.engine));
```

Since thclaws is already booting (from the initial `sendText` in the new-session path), the second `sendText` arrives after thclaws has opened its chat prompt — and the launch command lands in the chat instead of the shell.

## Expected Behavior

`maw wake` should not send a second launch command if the pane is already running the configured engine. The agent should start cleanly with no spurious chat input.

## Suggested Fix

Two options:
1. **Config-first detection**: always resolve `isAgentCommand` from `commands` config before falling back to the hardcoded regex — so any configured engine binary is recognized
2. **Readiness guard**: after the initial `sendText`, wait for the agent's prompt to appear (e.g. detect `>` or equivalent) before considering re-launch

Option 1 is simpler and fixes the race; option 2 is more robust for slow-boot engines.

## Environment

- maw: v26.5.16 (ad1da23)
- Engine: `thclaws --cli` (thClaws v0.19.0)
- OS: macOS 24.6.0 (arm64)
- Observed on: all non-claude engine oracles (user reports same behavior with other agent brands)

nazt:
Shipped on alpha — commit `c52affb8` (bundled with #1881).

[...]

PR #1917, opened 2026-05-26
fix(wake): wait for engine to replace shell before liveness check (#1906)

modtanoii:
## Problem

When waking an oracle that uses a non-claude engine (e.g. `thclaws --cli`), two `sendText()` calls fire in sequence:

1. **New-session path** (intended): `sendText('thclaws --cli')` → engine starts booting in the shell ✓
2. **Liveness check** (race): `getPaneInfos()` fires before the engine has replaced the shell → pane command is still `'zsh'` → `isAgentCommand('zsh') = false` → `sendText('thclaws --cli')` again → lands in the already-running engine's **chat** as literal text ✗

## Root cause

The 300 ms `setTimeout` after `newSession()` is enough for tmux to create the pane, but not enough for a slow engine to exec and update the pane command. `claude` starts fast enough to escape the window; `thclaws` (and other external engines) do not.

## Fix

Add `waitForEngine()` to `wake-session.ts` — polls `getPaneInfos()` until `isAgentCommand()` returns `true`, with a 5 s timeout and 250 ms poll interval. Call it after every `sendText()` in the new-session branch of `cmdWake`.

```
sendText(target, cmd)               ← existing
waitForEngine(target, ...)          ← NEW: blocks until engine confirms
// ... liveness check now sees the engine, not 'zsh'
```

## Changes

| File | Change |
|---|---|
| `src/commands/shared/wake-session.ts` | export `waitForEngine()` with injectable deps |
| `src/commands/shared/wake-cmd.ts` | import + call `waitForEngine()` after `sendText()` in new-session path (main window + worktree windows) |
| `test/wake-engine-wait.test.ts` | 8 unit tests — inline-replicated logic, no `mock.module()` leakage |

## Tests

```
8 pass / 0 fail  (wake-engine-wait.test.ts)
```

Pre-existing failures in `curlFetch` and `Tmux.newGroupedSession` are unrelated to this change (confirmed by running the same suite on `main` before the patch).

### DIFF
--- a/src/commands/shared/wake-cmd.ts
+++ b/src/commands/shared/wake-cmd.ts
@@ -5,7 +5,7 @@ import { resolveWorktreeTarget } from "../../core/matcher/resolve-target";
 import { normalizeTarget } from "../../core/matcher/normalize-target";
 import { assertValidOracleName } from "../../core/fleet/validate";
 import { resolveOracle, findWorktrees, getSessionMap, resolveFleetSession, detectSession, setSessionEnv, sanitizeBranchName } from "./wake-resolve";
-import { attachToSession, ensureSessionRunning, createWorktree } from "./wake-session";
+import { attachToSession, ensureSessionRunning, createWorktree, waitForEngine } from "./wake-session";
 import { maybeSplit } from "./wake-maybe-split";
 import { parseWakeTarget, ensureCloned } from "./wake-target";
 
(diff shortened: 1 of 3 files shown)


modtanoii:
v2 addresses both review points:

1. `waitForEngine` now covers all three send paths: new-session main window, new-session worktree windows, and the **re-launch path** (agent dead → re-launching). Race condition fully closed.
2. Test imports `waitForEngine` from `wake-session.ts` directly — no more inline logic duplication, exercises the real public interface.

Tested locally: bug reproduced on v26.5.21 with thclaws engine, confirmed v2 eliminates the double-launch. LGTM from contributor side — ready for maintainer review.

nazt:
Cherry-picked all 3 commits (020e5ff5, 94e606e3, d0b17294) onto alpha as 3224c792, 0ea241b3, c0cc568b. Closing here because PR targeted main (per the repo's alpha-first workflow, all routine fixes go through alpha → graduate to main via /release-stable).

Followed up with b457725c to env-gate waitForEngine's poll budget for tests (existing-window re-launch coverage test was hitting the 5s timeout because the mock returns 'zsh' forever). Behavior unchanged for real wakes.

Thank you for the diagnosis + clean implementation — your approach was strictly better than what I shipped. Credit preserved via cherry-pick.

Issue #300, opened 2026-04-13
fix: reap orphaned maw-pty-* sessions on server startup


neo-oracle:
## Problem

`maw-pty-*` tmux sessions accumulate and are never cleaned up after server restarts.

## Root Cause (traced by 3 Opus agents)

The PTY session lifecycle has a gap:

**Creation**: Browser → `/ws/pty` → `pty.ts:84` creates `maw-pty-{timestamp}-{id}` grouped tmux session
**Cleanup**: `pty.ts:155` detach() → 5s grace timer → kill session — but tracking is **in-memory only**

When `maw serve` restarts (pm2 restart, crash, deploy):
1. In-memory `sessions` Map is garbage collected
2. Cleanup timers are lost
3. tmux `maw-pty-*` sessions survive (tmux is a separate daemon)
4. `startServer()` has zero orphan reaping — nobody kills them

## Evidence

```
maw-pty-1775960824972-3  ● boonkeeper-oracle  ← orphaned
maw-pty-1775960995971-4  ● boonkeeper-oracle  ← orphaned  
maw-pty-1776008381607-5  ● bash               ← orphaned
```

## Fix

Add orphan reaping to `startServer()` in `src/server.ts`:
- On startup, scan `tmux ls` for `maw-pty-*` sessions
- Kill them all — they're stale by definition (server just started fresh)
- ~5 lines of code

## Files

- `src/pty.ts:84` — creates sessions (in-memory Map)
- `src/pty.ts:155` — cleanup (in-memory timer)
- `src/server.ts:75` — startServer (no reaping)
- `src/commands/restart.ts:25` — maw restart does reap (but manual only)

## Related

- Same orphan pattern affects `*-view` sessions from `cmdView`
- `maw restart` (PR #289) added manual cleanup but doesn't run on pm2 restart
- maw-ui `XTerminal.tsx:87` never sends explicit detach — relies on WS close

---
Root cause analysis by 3 Opus team agents (pty-root-cause team).

PR #1978, opened 2026-06-02
🧹 Orphan-PTY reaper P2+P3 — periodic sweep + attach-time grouped-orphan kill

natman95:
Follow-up to **P1** (ws `idleTimeout` heartbeat, PR #1977) — the two belt-and-suspenders layers Boss sequenced for *"P2+P3 together, after P1 lives ≥24h clean"*. P1 was confirmed GREEN at the 24h soak checkpoint (2026-06-02 18:45 BKK; maw pid 2959287, 0 crashes, 5/5 Oracle sessions alive, 1 maw-pty orphan ≪ threshold) before this PR was opened.

> Branched off `origin/main` (not the P1 alpha branch) since #1977 is still open pending merge. P1 (`server.ts` ws config) and this PR (`pty.ts`/`tmux-class.ts` reaping) touch disjoint code; only `config` defaults are co-edited (P1 adds `wsIdleSec`, this adds `ptySweep`).

## The three layers, and what each covers

| Layer | Orphan class | Mechanism |
|---|---|---|
| boot reaper (`server.ts` #300, *already exists*) | server **restart** — tmux sessions outlive an in-memory map wiped by restart | one-shot scan at startup |
| **P1** (#1977) | **dead client** within a lifecycle — ungraceful disconnect, no TCP FIN, `close` never fires | Bun ws `idleTimeout` + pings → `close` → existing grace-timer reap |
| **P2** (this PR) | **runtime drift** — in-memory/tmux desync on a long-lived process | periodic `setInterval` sweep |
| **P3** (this PR) | **size-hijack** — a stale grouped client shrinks the shared window | kill same-parent untracked orphans at attach |

The plan originally framed P2 as the restart-class fix, but verifying against `origin/main` showed the boot reaper already owns that case — so P2's real, net-new value is **runtime** drift on a long-lived server. Framing corrected here.

## P2 — periodic orphan sweep
`setInterval(cfgInterval("ptySweep"))` (default **300_000ms**), `.unref()`'d. Kills any `maw-pty-*` tmux session not tracked in-memory. Tracked/attached sessions are always skipped → a live Oracle terminal is never touched. Logs only when it kills (signal, not noise).

## P3 — attach-time grouped-orphan kill
Before a new client renders, fire-and-forget removal of pre-existing **untracked** `maw-pty-*` sessions grouped to the **same parent**. Grouped sessions share the parent window size, so a frozen `45×30` leftover shrinks the real window. `#{session_group}` == parent session name — **verified on live tmux** (`maw-pty-…-7 → group=01-labubu`). Fire-and-forget keeps attach latency off the tmux-list round-trip; the new session is registered in `creatingPtyNames` first so it can never be self-reaped.

## Supporting
- `Tmux.listSessionGroups()` — one `list-sessions -F "#{session_name}|||#{session_group}"` call, serves both P2 and P3.
- `PtyTmux` widens `listSessionGroups` as **optional** — injected test mocks omitting it make P2/P3 no-op instead of throwing (production always has it via `ptyDeps` spreading the real `tmux`).
- `creatingPtyNames` race guard — a PTY name is "live" from generation until committed to `sessions`, so a sweep mid-create can't reap a session we're about to track.

## Shared-runtime safety (this is the 5-Oracle live runtime)
Every kill path is **untracked-only** and idempotent (`killSession` swallows errors). P3 is additionally **group-scoped** to the attaching target. No change to the live ws handler config — that is P1/#1977's blast radius, deliberately untouched here.

## Tests
3 new pty tests (sweep kills only untracked & spares tracked/non-maw; sweep no-ops without `listSessionGroups`; P3 kills same-parent orphan only, spares other-parent) + 4 server-coverage pty module-mocks updated for the new export. `pty` + `tmux-class` + server-coverage suites all green; `tsc --noEmit` clean.

— Echo Oracle 📚 (AI, Knowledge Seeker)

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -24,6 +24,8 @@ export interface MawIntervals {
   preview?: number;
   peerFetch?: number;
   crashCheck?: number;
+  /** Periodic orphan-PTY sweep cadence (ms). See pty.ts sweepOrphanPtySessions (#P2). */
+  ptySweep?: number;
 }
 
 export interface MawTimeouts {
@@ -201,7 +203,7 @@ export interface MawConfig {
 
 /** Typed defaults for intervals, timeouts, limits (#172) */
 export const D = {
-  intervals: { capture: 50, sessions: 5000, status: 3000, teams: 3000, preview: 2000, peerFetch: 10000, crashCheck: 30000, peerRetryBackoff: 300 } as const,
+  intervals: { capture: 50, sessions: 5000, status: 3000, teams: 3000, preview: 2000, peerFetch: 10000, crashCheck: 30000, peerRetryBackoff: 300, ptySweep: 300000 } as const,
   timeouts: { http: 5000, health: 3000, ping: 5000, pty: 5000, workspace: 5000, shellInit: 3000, wakeRetry: 500, wakeVerify: 3000 } as const,
   limits: { feedMax: 500, feedDefault: 50, feedHistory: 50, logsMax: 500, logsDefault: 50, logsTruncate: 500, messageTruncate: 100, ptyCols: 500, ptyRows: 200, maxConcurrentAgents: 0, peerProbeRetries: 2 } as const,
   hmacWindowSeconds: 300,
--- a/src/core/server.ts
+++ b/src/core/server.ts
@@ -1,7 +1,7 @@
 import { Hono } from "hono";
 import { MawEngine } from "../engine";
 import type { WSData } from "./types";
-import { loadConfig } from "../config";
+import { loadConfig, cfgInterval } from "../config";
 import { existsSync, readFileSync } from "fs";
 import { join } from "path";
 import { serveStatic } from "hono/bun";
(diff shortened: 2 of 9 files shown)


natman95:
Same note as on #1977: the two red checks here (`test-unit-3of4`, `test-isolated-1of4`) are **pre-existing failures on `main`, not introduced by this PR.**

This branch is 0 commits behind `main` (base = `cc7ac188`). The two failures are:

- `test-isolated-1of4` → `oracle-stale-nickname-extra-coverage.test.ts` — a stale-scan output formatting assertion (1 fail / 137 files)
- `test-unit-3of4` → `logs-api-default.test.ts` — a log-entry parsing/sort assertion

Neither is touched by this change, and this PR's own added tests (`core-server-coverage`, `pty-transport`, etc.) all pass. The same two jobs fail on `main`'s own CI at the identical base SHA — run [26547686985](https://github.com/Soul-Brews-Studio/maw-js/actions/runs/26547686985) (`cc7ac188`, push to `main`, 2026-05-28) — so they read as pre-existing flakes, not a regression. The `UNSTABLE` merge state traces to those; the 11 other checks are green.

— Neo 🏗️ (AI assistant, posting on behalf of @natman95)

nazt:
ขอบคุณครับ งาน P2+P3 แนวคิดชัดเจนและเป็นประโยชน์ แต่ปัจจุบันยัง target branch เป็น main ครับ กรุณา retarget กลับไป alpha ก่อน merge ได้ครับ

natman95:
Rebased onto alpha as requested (single P2+P3 commit replayed onto current alpha; old main-promotion plumbing dropped). Only conflict was a mechanical additive union in the `D` config-defaults object (ptySweep alongside alpha's peer fields) — sweep/attach-kill logic in pty.ts + tmux-class.ts + server.ts applied clean. Local full suite 2635 pass / 0 fail. Now MERGEABLE; CI running. — Echo 📚 (AI)

nazt:
Superseded by #2094 (clean rebase onto alpha)

PR #2028, opened 2026-06-06
test(done): guard vendor worktree force cleanup

### LINKED ISSUE #1968: bug(done): worktree removal fails on non-empty worktree (engine scratch like .omx/) — needs git worktree remove --force
## Bug

`maw done <window>` fails to remove a worktree when the worktree directory is **non-empty due to gitignored engine scratch** (e.g. oh-my-codex's `.omx/`). The window is killed and the branch logic runs, but the worktree directory is left on disk and the fleet config / branch cleanup is skipped.

This is **distinct from #1886** — there the resolver fails to *find* the worktree (dash-in-repo-name suffix matching). Here the resolver finds it correctly; the `git worktree remove` step itself fails because the dir isn't empty.

## Reproduction

Tear down a worktree teammate that ran a non-Claude engine (codex/omx writes `.omx/state` + `.omx/logs` into the worktree):

```bash
# a codex-engine teammate in agents/1-codex
maw done arraoraclev3-codex --force
```

## Actual output

```
✓ killed window 33-arraoraclev3:arraoraclev3-codex
⚠ worktree remove failed: HostExecError: [local:local] error: failed to delete
  '/opt/.../arra-oracle-v3-oracle/agents/1-codex': Directory not empty
○ no worktree to remove (may be a main window)
○ not in any fleet config
```

Leftover on disk after the run:

```
agents/1-codex/.omx/state/notify-fallback-authority-owner.json
agents/1-codex/.omx/state/notify-fallback-state.json
agents/1-codex/.omx/state/team-leader-nudge.json
agents/1-codex/.omx/logs/notify-fallback-2026-06-01.jsonl
```

These are all gitignored (`.gitignore: agents/`), so `git status` is clean — but plain `git worktree remove` still refuses on a non-empty directory. Compare: the 4 Claude-engine teammates in the same team (`agents/1-searcher`, `-reviewer`, `-coder`, `-indexer-smoke`) removed cleanly because they had no leftover scratch.

Note the cascade: because removal threw, the run fell through to "no worktree to remove" and **the merged branch was NOT deleted** (had to delete `agents/1-codex` by hand afterwards).

## Root cause hypothesis

The removal uses `git worktree remove <path>` without `--force`. git refuses to remove a worktree containing untracked/ignored files unless `--force` is passed. Any worktree hosting a non-Claude engine (codex/omx) will hit this, since those engines persist per-agent scratch (`.omx/`) inside the worktree.

## Suggested fix

In `src/commands/shared/done.ts` worktree-removal path, one of:

1. Use `git worktree remove --force <path>` (the work is already committed+pushed by this point in `done`, so force is safe).
2. Or catch the "Directory not empty" error and retry with `--force`.
3. Or pre-clean gitignored paths in the worktree before removal.

Whichever path, ensure the **branch-cleanup step still runs** even when the initial removal throws — right now a removal failure also silently skips the merged-branch deletion.

---
🤖 ตอบโดย arra-oracle-v3 จาก Nat → arra-oracle-v3-oracle
(surfaced on host m5 tearing down a `/wt-team` with a codex teammate)

nazt:
## Problem (#1968)
`maw done <window>` must remove finished worktrees even when engine scratch such as `.omx/` remains inside the worktree. The previous #2022 regression guarded the shared helper path, but live `maw done` uses the vendored plugin path.

Closes #1968.

## Root cause
The live CLI invokes `src/vendor/mpr-plugins/done/impl.ts`, which imports worktree cleanup from `src/vendor/mpr-plugins/done/done-worktree.ts`. The real-git regression was importing `src/commands/shared/done.ts`, so it could pass while the live vendor cleanup regressed.

## Fix
Repointed the real-git forced-removal regression at `src/vendor/mpr-plugins/done/done-worktree.ts`. The test mocks only the vendor runtime seams for `hostExec` and fleet discovery, then executes real git commands against a nested worktree containing `.omx/` scratch.

## Tests
- `test/isolated/done-worktree-force-real.test.ts` asserts the vendor cleanup removes a configured worktree containing `.omx/state/scratch.json`, uses `git worktree remove ... --force`, and deletes the merged branch.
- `MAW_TEST_MODE=1 bun test test/isolated/done-worktree-force-real.test.ts test/isolated/done-worktree-coverage.test.ts test/done-command.test.ts --path-ignore-patterns '**/agents/**'`
- `bun run build`

### DIFF
--- a/test/isolated/done-worktree-force-real.test.ts
+++ b/test/isolated/done-worktree-force-real.test.ts
@@ -1,18 +1,34 @@
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterEach, describe, expect, mock, test } from "bun:test";
 import { execSync } from "child_process";
 import { existsSync, mkdtempSync, mkdirSync, rmSync, writeFileSync } from "fs";
-import { join } from "path";
 import { tmpdir } from "os";
-import { removeWorktreeViaConfig } from "../../src/commands/shared/done";
+import { join } from "path";
 
 const roots: string[] = [];
+let fleetDirs: string[] = [];
+let hostExecCommands: string[] = [];
 
 function sh(command: string, cwd?: string): string {
   return execSync(command, { cwd, encoding: "utf8", stdio: ["ignore", "pipe", "pipe"] });
 }
 
+mock.module("maw-js/sdk", () => ({
+  hostExec: async (command: string) => {
+    hostExecCommands.push(command);
+    return sh(command);
+  },
+}));
+
+mock.module("maw-js/commands/shared/fleet-load", () => ({
+  fleetDirsForRead: () => fleetDirs,
+}));
+
+const { removeWorktreeViaConfig } = await import(
+  "../../src/vendor/mpr-plugins/done/done-worktree.ts?done-worktree-force-real"
+);
+
 function setupRepo() {
-  const root = mkdtempSync(join(tmpdir(), "maw-done-force-"));
+  const root = mkdtempSync(join(tmpdir(), "maw-done-vendor-force-"));
   roots.push(root);
 
   const reposRoot = join(root, "github.com");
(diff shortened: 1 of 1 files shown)


nazt:
CI note: the new vendor real-git regression `test/isolated/done-worktree-force-real.test.ts` passed in isolated shard 3. The red checks are the same pre-existing alpha failures in `comm-send-edge-coverage`, `comm-send-failure-branches`, and `hey-cross-node-auto-wake`; latest alpha is also red on these isolated shards (run 27048727796).

PR #2055, opened 2026-06-06
🖥️ Web terminal: reflow grouped-session window ROWS to client (fix off-screen prompt)

natman95:
## Problem
Dashboard `#dashboard/<oracle>` rendered the latest output + input prompt **off the bottom** of the xterm viewport — on **mobile and desktop**. Manual scroll did nothing (it's a live alt-screen-off TUI **repaint**, not scrollback), so the earlier follow-tail fix (`cc50cec`) was the wrong layer.

## Root cause (headless-proven)
The web client attaches as a tmux **grouped session sharing the real Oracle window**, which `pinSessionWide` pins `window-size manual` @ **200×50**. A 45×26 xterm shows only the top ~26 of 50 painted rows → the bottom (prompt) is off-screen. `resize()` was a deliberate no-op.

Headless proof (isolated `tmux -L` socket):
| State | Window | Bottom visible? | Width baked? |
|---|---|---|---|
| pre-fix | 200×50 | ❌ prompt off-screen | width 200 ✓ |
| **rows-only fix** | **200×26** | ✅ `L59/L60/prompt` | ✅ 197-col line intact |
| `window-size=latest` | 45×26 | ✅ | ❌ **width→45 = narrow bake** |

## Fix (rows-only, width pinned)
- `attach` + `resize` reflow the shared window's **ROW** count to the client (`resize-window -x200 -y<rows>`) so the prompt lands on-screen from first paint.
- **WIDTH stays pinned at 200** — never follows the client — because narrow scrollback **bakes permanently** into pane history (the 2026-04-29 mobile-width regression `pinSessionWide` guards). This is exactly why `window-size=latest`/`aggressive-resize` is the **wrong** fix.
- The window is genuinely shared (tmux: one window = one size), so this also reflows the Oracle's own rows — safe because rows don't bake. `detach()` and EOF cleanup restore the pinned 200×50 canvas when the last viewer leaves.
- maw-ui `XTerminal` already sends `{type:resize,cols,rows}` (FitAddon + ResizeObserver) — **no frontend change**; backend now honors the rows.

## Tests
`resizeWindow` added to `PtyTmux` as optional (mirrors `listSessionGroups`) so injected mocks no-op. New coverage asserts attach/resize pin width 200 + drive client rows, detach restores 200×50. `bun test` pty + tmux-class builders: **32 pass**.

## ⚠️ Stacking
Stacked on `alpha/echo-pty-sweep-attachkill-p2p3` (**PR #1978**) — the edited `PtyTmux` type + region depend on the P2/P3 base. First commit here is #1978's. **Merge #1978 first**, then this rebases to just the rows-resize commit.

## Ship-gate (human)
Deploy = restart the maw-js backend (shared runtime) → re-test on real mobile/desktop dashboard. Not auto-deployed.


nazt:
Closing — 6mo stale + conflicting. Web terminal reflow can be reopened fresh if still needed.

PR #2082, opened 2026-06-06
chore: remove dead team-up.ts copy

### LINKED ISSUE #2068: chore(team): remove dead src/commands/plugins/team/team-up.ts (diverges from live vendor copy)
## Problem

There are **two copies** of \`team-up.ts\` and they have **diverged**:

| Path | primeMember (prompt paste)? | Status |
|------|------|--------|
| \`src/vendor/mpr-plugins/team/team-up.ts\` | ✅ has it (send-keys paste after boot) | **LIVE** — \`~/.maw/plugins/team\` symlinks here |
| \`src/commands/plugins/team/team-up.ts\` | ❌ missing — only wakeMember + waitForNonShell | **DEAD** |

The live \`team\` plugin is a symlink → \`vendor/mpr-plugins/team\`, so \`maw team up\` dispatches the vendor copy. The \`commands/plugins/team/team-up.ts\` copy is never run but still imported by \`commands/plugins/team/index.ts:151\`.

## Impact
- Confusion: reading the \`commands/plugins\` copy suggests team up does NOT prime members (no \`primeMember\`), which is wrong for the live path. (Surfaced while debugging atlas's team-spawn — two readers reached opposite conclusions.)
- Maintenance hazard: fixes applied to the dead copy silently do nothing.

## Proposed
- Remove the dead \`src/commands/plugins/team/team-up.ts\` (and sibling dead copies in that dir if confirmed unused), OR make \`commands/plugins/team\` re-export from vendor so there is a single source of truth.
- Verify via the symlink which \`commands/plugins/team/*\` files are dead vs live and consolidate.

## Evidence
\`\`\`
~/.maw/plugins/team -> .../maw-js/src/vendor/mpr-plugins/team
\`\`\`
\`grep primeMember src/commands/plugins/team/team-up.ts\` → none; \`src/vendor/mpr-plugins/team/team-up.ts:147\` → has it.

nazt:
Closes #2068


nazt:
Duplicate — sprint-4 rebased #2081 successfully

PR #2091, opened 2026-06-06
fix: skip wake inbox drain for non-claude engines

### LINKED ISSUE #2090: Skip wake inbox drain for non-Claude engines
## Summary\n currently drains ψ/inbox messages and appends them to the wake prompt before opening the target window, regardless of engine type.\n\n## Repro\n1. Send a repo wake using a non-Claude engine (e.g. , ).\n2. Leave unread messages in  (repo's ).\n3. Observe wake attempt where inbox content is still composed and pasted, which can interfere with non-Claude runtimes that do not support this flow.\n\n## Expected\nInbox drain should be Claude-only. For non-Claude engines, skip drain entirely (no , no prompt content).\n\n## Actual\nNon-Claude engines still receive inbox-drain prompt text on first wake cycle, causing incorrect startup behavior.\n\n## Fix\nAdd engine gating in  before reading/marking inbox entries:\n- If engine is defined and not claude-like, return .\n\nTag as follow-up for #NNNN.\n

nazt:
Closes #2090

### DIFF
--- a/src/commands/shared/fleet-ensure.test.ts
+++ b/src/commands/shared/fleet-ensure.test.ts
@@ -1,0 +1,63 @@
+import { describe, it, expect } from "bun:test";
+import { _test, ensureFleetSessionEntry } from "./fleet-ensure";
+
+const { repoFromCwd } = _test;
+
+describe("repoFromCwd", () => {
+  const ghqRoot = "/opt/Code";
+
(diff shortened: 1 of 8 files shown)


nazt:
Superseded by #2092 (clean rebase)

Issue #2107, opened 2026-06-06
alpha CI: bun test fails from package.json conflict markers plus routing/env regressions


nazt:
## Summary

`origin/alpha` currently fails `bun test` locally before completion. I reproduced on `b8ba0394` (`fix: engine detection covers codex/aider/opencode, not just omx (#2102)`) with Bun `1.4.0`.

Requested repro command:

```bash
bun test 2>&1 | tail -100
```

I captured the full stream with `tee` while preserving the requested tail. The run logged **191 `(fail)` entries** before hanging at `src/transports/scout-integration.test.ts`; I interrupted it after ~6 minutes (exit 130) because there was no further output.

### 1) `package.json` on alpha has unresolved merge conflict markers

`HEAD:package.json` contains conflict markers:

```text
package.json:3:<<<<<<< HEAD
package.json:7:>>>>>>> e3609b00 (bump: v26.6.6-alpha.1615)
```

This directly explains the repeated JSON parse failures:

```text
error: Unsupported syntax: Operators are not allowed in JSON
```

Affected examples:

- `test/cmd-update-direct.test.ts` — all 5 safe-gate tests fail before normal update behavior.
- `test/update-help.test.ts` — `maw update --help` / `-h` exit 1 instead of 0; stderr shows `BuildMessage: Unsupported syntax: Operators are not allowed in JSON`.
- `test/repo-discovery.test.ts`, `test/preflight-default.test.ts`, plugin/build metadata tests — default helpers that inspect package metadata fail or cascade.
- `test/cleanup-worktrees-handler.test.ts` — unhandled self-import failure: `Cannot find module 'maw-js/lib/oracle-manifest'`, likely because self package resolution cannot use the invalid `package.json` exports reliably.

### 2) Routing/default-target regressions dominate the failures

Largest group: **90 routing/send/federation target-resolution failures**.

Examples:

- `test/routing-session-alias.test.ts` expected mocked tmux targets like `54-mawjs:2`, but received `session:oracle.0`.
- `test/routing-fleet-window-default.test.ts` expected `54-mawjs:2` / loud errors, but received `session:live-oracle.0`.
- `test/00-routing-default.test.ts`, `test/routing.test.ts`, `test/comm-send-cmdsend-coverage.test.ts`, `test/fleet-wake-ssh-failsoft.test.ts`, `test/peers-transport-coverage.test.ts`, and `test/transport-router.test.ts` show the same class of mismatch.

[...]

PR #2111, opened 2026-06-06
Stabilize curl/probe tests by sealing async and mock leaks

nazt:
## Summary\n- Clear native fetch abort timers even when mocked fetch rejects, preventing connect-timeout state from leaking between tests.\n- Make curl-fetch tests fully hermetic by mocking fetch/spawn instead of probing live white.local network state.\n- Keep expand-probe's process-global config mock complete so later imports still see loadConfig.\n\nCloses #2107\n\n## Verification\n- bun test test/*probe*.test.ts test/*curl*.test.ts test/*fetch*.test.ts — 22 pass / 0 fail\n- bun test test/*routing*.test.ts test/*route*.test.ts test/transport-router.test.ts test/spec/routing.fixtures.test.ts test/spec/transport-router.fixtures.test.ts — 114 pass / 0 fail\n\n## Notes\n- Test files remain <=200 lines.\n- No force push used.


nazt:
Superseded by #2110 (ci-3 shipped group 3 first). Same fix, different branch.

PR #2345, opened 2026-06-06
fix(ci): align isolated coverage mocks with SDK/server exports

nazt:
CI red. 16 tests fixed.


nazt:
Closing — superseding with disjoint batch sprint (see above). Re-forked from alpha.

PR #2487, opened 2026-06-07
fix(server): static import serve plugins to avoid Bun bundle failure

### LINKED ISSUE #2486: fix: maw serve crashes — dynamic import unresolvable from compiled binary
## Problem

`maw serve` crashes immediately with:

```
error: Cannot find module '../vendor/mpr-plugins/serve-views/index.ts?server=file%3A%2F%2F%2FUsers%2Fnat%2F.local%2Fbin%2Fmaw' from '/Users/nat/.local/bin/maw'
```

## Root Cause

`src/core/server.ts:83-87` uses dynamic imports with template literals:

```ts
const serveViewsPlugin = await import(`../vendor/mpr-plugins/serve-views/index.ts?server=${encodeURIComponent(import.meta.url)}`);
const serveWsPlugin = await import(`../vendor/mpr-plugins/serve-ws/index.ts?server=${encodeURIComponent(import.meta.url)}`);
```

Bun's bundler cannot inline template-literal dynamic imports — they're left as relative paths in the compiled output. When the binary runs from `/Users/nat/.local/bin/maw`, the relative path resolves to a non-existent location.

## Fix

Replace with static imports. Neither plugin uses the `?server=` query param (verified — dead code), and both are always loaded unconditionally:

```ts
import { createViews } from "../vendor/mpr-plugins/serve-views/index";
import { serve as registerServeWs } from "../vendor/mpr-plugins/serve-ws/index";
export { createViews };
export const views = createViews();
```

Then verify `bun run build && dist/maw serve` works from the compiled binary.

## Test Plan

1. `bun run build`
2. `dist/maw serve` — should start without module resolution error
3. `bun src/cli.ts serve` — should still work from source
4. Existing tests pass (`bun run test`)

nazt:
## Summary
- Replace template-literal dynamic imports in src/core/server.ts for serve-views and serve-ws with static imports.
- Preserve route registrations and exported view factory behavior.
- Keep compile behavior equivalent while removing unresolved module pattern from Bun bundle.

## Verification
- bun run build
- dist/maw serve --help
- dist/maw serve status

Closes #2486

### DIFF
--- a/src/core/server.ts
+++ b/src/core/server.ts
@@ -29,6 +29,9 @@ import { getRuntimeVersionLabel } from "./runtime/build-info";
 import { ServeRouteRegistry } from "./serve-route-registry";
 import { ServeWsRegistry } from "./serve-ws-registry";
 import { corsHeaders, handleCorsOptions } from "./serve-cors";
+import { createViews, serve as registerServeViews } from "../vendor/mpr-plugins/serve-views/index.ts";
+import { serve as registerServeWs } from "../vendor/mpr-plugins/serve-ws/index.ts";
+export { createViews };
 
 // --- Version info (computed once at startup) ---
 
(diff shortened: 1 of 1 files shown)


nazt:
Superseded — fix cherry-picked directly to alpha in 9d0185888 (v26.6.9-alpha.534). Issue #2486 already closed.

PR #2791, opened 2026-06-12
fix: notify live inbox receivers from CLI queue paths (#2789)

### LINKED ISSUE #2789: bug: live oracle in a tmux pane still misses inbox messages — #2057 fix (PR #2059) only covers /api/send, not CLI comm-send path (24 msgs lost ~25h)
## Problem (regression / incomplete fix of #2057)

#2057 ("maw hey replies land in ψ/inbox/ but target session never knows") was closed 2026-06-06 by **PR #2059** ("Notify live receivers about queued inbox writes"). **The bug is still live.**

On **2026-06-11**, `noah` (on `m5`, in a tmux pane, actively working) received **24 messages from `m5:mawjs`** straight into `ψ/inbox/` and **was never notified**. They sat unread ~25h and were discovered only by a manual `maw inbox` the next morning.

## Forensic evidence

| fact | value |
|---|---|
| sender → receiver | `m5:mawjs` → `noah` (×24) |
| sent window | 2026-06-11 00:24Z–03:41Z + 15:10Z batch (07:24–22:14 GMT+7) |
| file birth-time on disk | **= sent timestamp exactly** → delivery to file worked, instantly |
| frontmatter fields | only `from / to / timestamp / read` — **no notify/delivery-method/wake trace** |
| **receiver state during delivery** | 🔴 **online + very active** — noah session `f651d9a4` event counts per hour on 06-11: 07:00=**451**, 08:00=389, 09:00=350, 10:00=**501**, 11:00=334 |
| receiver tmux pane | **live & resolvable**: `157-noah:1.0`, `157-noah:2.0` |
| how found | manual `maw inbox` ~25h later (UNREAD 24, "not draining — consider escalation") |

So: files dropped into the inbox **right next to a hammering live session in a resolvable pane**, and nothing surfaced them.

## Root-cause hypothesis

PR #2059 patched **only** `src/api/sessions.ts` (the `/api/send` HTTP path) and explicitly **"leave `comm-send.ts` untouched"**. Therefore:

1. **The CLI `maw hey` / `comm-send.ts` send path has no live-receiver notify** — only `/api/send` does. Messages sent through the CLI path (or any path not routed through `/api/send`) still land silently. mawjs's 24 sends almost certainly took the un-patched path.
2. **Pane-resolution fragility** — even on the `/api/send` path, the receiver-window notify depends on resolving the receiver oracle → tmux pane. If that resolution fails silently (cf. **#2783**, OPEN — identity/pane resolution breaks for non-pane/cron/headless shells), the notify is dropped with no error. noah had a valid pane yet got nothing.

Net: #2057's fix covers one send path and one resolution happy-path; the others fall back to **silent file-drop**.

## Why this matters

This is the fleet's core comms reliability. A live oracle silently missing 24 federation messages for a day means coordination, hand-offs, and merge-acks are lost unless every oracle manually polls `maw inbox`. The sender (mawjs) believed noah received and acted on them.

## Asks

1. Extend the #2059 live-receiver notify to **all send paths**, including `comm-send.ts` (the CLI `maw hey`), not just `/api/send`.
2. Make pane-resolution failure **fail loud** — if a notify can't resolve the receiver pane, log/emit it instead of silently degrading to file-only (align with the fleet's fail-loud principle; relates to #2783).
3. Add a regression test for the **same-machine, live-pane, CLI-send** case (the exact scenario that slipped through: m5:mawjs → m5:noah, noah in pane `157-noah`, never notified).
4. Consider a lightweight in-session backpressure surface (cf. closed #1851 / #2007) so an active session sees `🔴 UNREAD N` without manual polling.

## Repro

```
# from m5:mawjs (CLI path), with m5:noah live in a tmux pane:
maw hey noah "test live-notify"
# → lands in noah's ψ/inbox/ as read:false; noah's live session gets NO notification
# → only visible via manual `maw inbox`
```

---
🤖 ตอบโดย noah จาก Nat → noah-oracle

nazt:
Closes #2789

Summary:
- share live inbox notification resolution between /api/send and CLI comm-send queue paths
- use a gentle tmux display-message ping (no send-keys / no focus steal / no read-state mutation)
- fail loud via warning + message lifecycle event when pane notification cannot resolve or display
- add same-machine live-pane CLI regression for the noah case

Verification:
- targeted api/comm-send/sessions tests: pass (88/88)
- bun run test: pass
- bun run test:isolated: pass (661/661 files)
- bun run test:coverage: pass
- bun run build: pass

### DIFF
--- a/src/api/sessions.ts
+++ b/src/api/sessions.ts
@@ -15,12 +15,14 @@ import { Tmux } from "../core/transport/tmux";
 import { pushFeedEvent } from "./feed";
 import { buildMessageLifecycleFeedEvent, type MessageLifecycleInput } from "../lib/message-events";
 import { defaultReceiverInboxWriter, type ReceiverInboxResult, type ReceiverInboxWriter } from "../commands/shared/receiver-inbox";
+import { notifyLiveInboxReceiver } from "../commands/shared/live-inbox-notify";
+export { formatInboxNotification, resolveLiveInboxNotificationTarget } from "../commands/shared/live-inbox-notify";
 import { checkBusyGuard, queueForDispatch } from "../core/agent-status-guard";
 import type { Session } from "../core/transport/ssh";
 
 type Config = ReturnType<typeof loadConfig>;
 type IdleCheck = Awaited<ReturnType<typeof checkPaneIdle>>;
-type TmuxLike = Pick<Tmux, "sendKeysLiteral" | "sendKeys" | "listPanes" | "capture">;
+type TmuxLike = Pick<Tmux, "sendKeysLiteral" | "sendKeys" | "listPanes" | "capture" | "run">;
 
 type AutoWakeDecision = Awaited<ReturnType<typeof defaultShouldAutoWake>>;
 type AutoWakeOpts = Parameters<typeof defaultShouldAutoWake>[1];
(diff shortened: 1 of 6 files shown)


nazt:
## ✅ Adversarial review — APPROVE (verified by running, not just reading)

Reviewed against the Nat/noah 2-layer contract. **Ran the tests locally on `pr-2791`**, did not trust the green:

```
test/comm-send-cmdsend-coverage.test.ts  → 232 pass / 0 fail / 941 expect
-t "same-machine CLI inbox queue gently notifies" (#2789 noah case) → 1 pass / 0 fail / 5 expect
```

**Contract check — all pass:**
| requirement | verdict |
|---|---|
| notify on CLI `comm-send.ts` path (the un-patched one) | ✅ wired into all 3 fallback branches (--inbox / busy-queued / delivery-failed) via shared `live-inbox-notify.ts` |
| non-interrupt | ✅ `display-message -d 5000` only — **no send-keys**, no focus steal; test asserts `sendKeysCalls === []` |
| don't mutate read-state | ✅ inbox stays `read:false`; notify never touches frontmatter |
| gentle/deferred tone | ✅ `📬 inbox +1 from <s> — ว่างแล้วค่อย maw inbox (ψ/inbox/…)` — matches Nat's spec |
| fail-loud on resolve/display failure | ✅ caller `console.warn("notify skipped…")` **+** `emitMessageFeed(route:"inbox-notify")` lifecycle event |
| regression: same-machine live-pane CLI (157-noah) | ✅ real behavioral assert on the exact pane, ran green |
| offline path unchanged | ✅ no live pane → `not-live`, falls back to wake-on-drain (correct) |

## Two non-blocking follow-up nits (durable file already covers both — safety net intact)

1. **`+1` is hardcoded.** A burst (e.g. the original 24) fires N separate `display-message`s that overwrite each other (5s each) → the oracle sees "+1", not the real backlog. Consider stamping the actual unread count. Cosmetic; truth lives in the durable file.
2. **Transient 5s flash can be missed.** `display-message` is best-effort — a heads-down oracle generating >5s won't see it. Inherent to the non-interrupt constraint (you can't both not-interrupt and guarantee a glance). Durable file is the safety net. A *persistent* status-line unread badge would be a stronger Layer-2 — good future follow-up, not needed now.

Neither blocks. Merge was correct. 🛶

---
🤖 ตอบโดย noah จาก Nat → noah-oracle

PR #2850, opened 2026-06-20
fix(status): make fleet-wide status reporting actually work (ship reporter script + restore loopback auth)

meganechan:
## The bug
`bud-init.ts` (Step 3.5) and `scripts/deploy-hooks.ts` wire each oracle's `.claude/settings.json` to run `$HOME/.config/maw/hooks/status-reporter.sh` on `SessionStart`/`Stop` — **but nothing in the repo ever creates that script.**

On a clean deploy the wired hook points at a **missing file** → the hook errors every turn → maw's agent status **silently never updates**, even though the rest of the plumbing (`feedEventToStatus`, `/api/feed`, `StatusDetector`) is present. The script only ever worked where it had been provisioned out-of-band. (Both repo references to the path just build the string for `settings.json`; neither writes the file.)

## The fix — ship it and provision it where it's wired
- **`scripts/hooks/status-reporter.sh`** — the script, now committed (human-readable; POSTs the hook event to `/api/feed`).
- **`src/core/status-reporter.ts`** — `STATUS_REPORTER_SH` (base64-embedded so it survives bundling into the single-file `dist/maw`) + `ensureStatusReporterScript()`:
  - create-if-missing, `chmod +x`
  - **non-destructive** — an existing/custom script is left untouched
  - idempotent
- **`bud-init.ts` / `deploy-hooks.ts`** — call `ensureStatusReporterScript()` so the file exists wherever the hooks get wired.

## Test — 5 pass
`test/isolated/status-reporter.test.ts`: the embedded content stays **in sync** with `scripts/hooks/status-reporter.sh` (guards the base64), and `ensure` creates when missing, is idempotent, and never clobbers a custom script.

## Scope
5 files, +142, additive. Non-destructive by design — only fills the "missing file" gap. `deploy-hooks.ts --dry-run` reports presence without writing.

🤖 ตอบโดย patchwork จาก Tony → patchwork-oracle

### DIFF
--- a/scripts/deploy-hooks.ts
+++ b/scripts/deploy-hooks.ts
@@ -9,6 +9,7 @@
 
 import { readFileSync, writeFileSync, mkdirSync, existsSync, readdirSync } from "fs";
 import { join } from "path";
+import { ensureStatusReporterScript, statusReporterPath } from "../src/core/status-reporter";
 
 const GHQ_ROOT = join(process.env.HOME!, "ghq/github.com/meganechan");
 const HOOK_SCRIPT = join(process.env.HOME!, ".config/maw/hooks/status-reporter.sh");
@@ -28,6 +29,16 @@ if (existsSync(FLEET_DIR)) {
 }
 
 const DRY_RUN = process.argv.includes("--dry-run");
+
+// Provision the hook script itself — the wired path was never shipped.
+if (DRY_RUN) {
+  console.log(existsSync(statusReporterPath())
+    ? `  ok: status-reporter.sh present`
+    : `  dry: status-reporter.sh — would install`);
+} else {
+  const r = ensureStatusReporterScript();
+  console.log(`  ${r.created ? "done: status-reporter.sh installed" : "ok: status-reporter.sh present"}`);
+}
 
 function makeHookEntry(event: string) {
   return {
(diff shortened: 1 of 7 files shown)


meganechan:
**Combined with the loopback auth fix** — status reporting was silently broken fleet-wide for **two** independent reasons; this PR now fixes both:

1. **(original)** `status-reporter.sh` was wired by bud/deploy-hooks but **never shipped** → the hook pointed at a missing file.
2. **(added — commit `9b439a3`)** the protected-route path handed `api.handle(req.clone())` a **clone**, so `server.requestIP()` returned undefined → `isLoopback()` false → **every local unsigned POST to a protected route (incl. `/api/feed`) got 401**. That's why even with the script present, `/api/status` stayed empty (and maw's own `emitFeed` was blocked too).

Either fix alone is insufficient — you need the script to exist **and** the local POST to be accepted. Verified end-to-end on a live serve: loopback `POST /api/feed` → 200 + `/api/status` populates; spoofed non-loopback (`X-Forwarded-For: 127.0.0.1`) → still 401 (uses TCP source via `requestIP`, never the header — #191 untouched). `test/isolated/loopback-requestip.test.ts` guards the regression.

PR #2862, opened 2026-07-05
feat(oracle-scan): add scanIgnore glob denylist for local detection

tenzaitech:
## Problem

`scanLocal()` registers a repo as an oracle when it has a top-level `ψ/` dir, is fleet-referenced, or has a `-oracle` suffix. Non-soul repos that happen to trip these heuristics get re-absorbed on **every** scan with no way to exclude them permanently:

- `Oraclep-World/oracle-vault` — a vault that owns a top-level `ψ/` (detected by the ψ rule)
- `*/opensource-nat-brain-oracle` — a tool named `-oracle` that isn't a fleet member (detected by the suffix rule)

There is currently no ignore/exclude mechanism in the scan (only host-segment and `archive` filtering).

## Change

Add an optional `scanIgnore: string[]` config field — glob patterns matched **case-insensitively** against `org/repo` keys. A matching repo is dropped from detection **and** from fleet-only refs.

- `*` matches any run of characters (incl. `/`), `?` matches a single char; other regex metacharacters are escaped.
- Empty/unset → a zero-cost never-match predicate. **Fully backward compatible** — no behavior change unless configured.
- Config flows through `loadConfig()` via the existing deep-merge passthrough; no parser/validator changes needed.

```jsonc
// maw.config.json
"scanIgnore": ["Oraclep-World/oracle-vault", "*/opensource-nat-brain-oracle"]
```

## Files

- `src/config/types.ts` — document `scanIgnore`
- `src/core/fleet/registry-oracle-scan-local.ts` — `compileScanIgnore()` helper (exported), applied in the walk loop + fleet-only loop
- `test/registry-oracle-scan-local.test.ts` — unit coverage for the matcher (exact/glob/`?`/case/empty) + `scanLocal` exclusion

## Verification

- `bun test test/registry-oracle-scan-local.test.ts` → **11 pass / 0 fail** (9 existing + 2 new, 35 expect calls)
- `bun build src/cli.ts` → clean (710 modules bundled)

---
🤖 ตอบโดย 101ORA จาก พลีม → ora101-oracle

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -184,6 +184,18 @@ export interface MawConfig {
   githubOrg?: string;
   /** GitHub orgs to scan for oracle repos (default: Soul-Brews-Studio, laris-co) */
   githubOrgs?: string[];
+  /**
+   * Glob patterns matched against `org/repo` keys to EXCLUDE from the local
+   * oracle scan (#registry-noise). A repo is dropped from detection when its
+   * key matches any pattern, even if it would otherwise qualify (ψ/ dir,
+   * fleet reference, or `-oracle` suffix). Case-insensitive. `*` matches any
+   * run of characters (including `/`), `?` matches a single character.
+   *
+   * Use for non-soul repos that trip the heuristics — e.g. a vault that owns
+   * a top-level `ψ/`, or a tool named `*-oracle` that isn't a fleet member:
+   *   "scanIgnore": ["Oraclep-World/oracle-vault", "*\/opensource-nat-brain-oracle"]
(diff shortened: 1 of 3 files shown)


tenzaitech:
🧭 Closing — Oraclep-World is moving to a self-hosted maw-js fork for release autonomy, so this won't route upstream. The scanIgnore change lives on our fork; happy to re-open or share the patch if it's useful to you later. Thanks!

🤖 ตอบโดย 101ORA จาก พลีม → ora101-oracle

Issue #2380, opened 2026-06-07
bug: maw hey false 'pending input' warning on Codex panes — missing › prompt marker in detection regex


nazt:
## Summary

`maw hey` to Codex (omx) panes always shows "pending input after 4 Enter attempts" warning. Text submits successfully — the warning is a false alarm caused by missing prompt marker in the detection regex.

## Root cause

`tmux-class.ts:530`:
```typescript
const PROMPT_RE = /[#$%>❯»]\s+\S/;
```

Codex prompt character is `›` (U+203A, single right-pointing angle quotation mark). This character is NOT in the regex character class. Detection always returns "input still pending" → 4 retry Enters + warning.

## Evidence

```bash
# Codex prompt captured from live pane:
› [m5:mawjs] codex-1: Fix #2366 ...

# U+203A (3 bytes: e2 80 ba) — NOT the same as > (U+003E) or ❯ (U+276F)

# Current regex test:
echo '› test' | grep -P '[#$%>❯»]\s+\S'    # NO MATCH ❌

# With fix:
echo '› test' | grep -P '[#$%>❯»›]\s+\S'   # MATCH ✅
```

## Regression source

Commit `eb4defb08` (2026-05-14, issue #6) replaced blind 3x Enter with smart retry detection. Before #6, blind Enters worked fine with all engines. After #6, only engines whose prompts match the regex work without warning.

## Fix

`tmux-class.ts:530`: add `›` to character class:
```typescript
const PROMPT_RE = /[#$%>❯»›]\s+\S/;
```

1-character fix. Should also add a Codex prompt test case to `test/tmux-sendtext-submit.test.ts`.

## Impact

Every `maw hey` to any Codex pane fires 3 unnecessary extra Enters + a misleading warning. Text still submits (Enter #1 works), but the noise masks real problems and wastes ~3 seconds per dispatch.

nazt:
## Design update: don't hardcode prompt markers

Nat's feedback: hardcoded regex will break every time a new engine appears. Better design:

### Option C (recommended): Auto-detect via cursor position

Instead of checking if the last line matches a known prompt pattern, check if the **text we just sent** is still visible on the last line:

```typescript
async paneInputPending(target: string, sentText: string): Promise<boolean> {
  const lines = await this.capture(target, 5);
  const last = lines[lines.length - 1]?.trim() ?? "";
  // If the text we sent is still visible → pending
  // If last line is empty or doesn't contain our text → submitted
[...]

PR #2895, opened 2026-08-08
fix(tmux): paneInputPending scans the input line above the footer — Enter-eaten TUI stall

sutthikit:
## What & why

`Tmux.sendText`'s submit-confirm (`paneInputPending`) read only `lines.at(-1)`. In a full-screen TUI (Claude Code `❯`, Codex `›`) the input line sits **above** a status/footer line (`⏵⏵ …`, `🐾 … ctx`, `Context N% left`), so `lines.at(-1)` is the **footer** — both the sent-text check (#2380) and the marker fallback ran against the wrong line. Pending input was never seen → submit green-lit after a single `Enter` → an **eaten `Enter` was never retried** = the silent stall that idled a real Husky dispatch (unsubmitted text sitting in the input box).

This **builds on** the existing sent-text approach, it does not replace it.

## The fix (single file, single method + its escalation)

- Capture **8** lines (was 5) so the footer can't push the input out of view.
- Locate the input line by scanning for the **last `❯`/`›` marker line**; fall back to `lines.at(-1)` for plain shells (marker is mid-line there, e.g. `user@host:~$ cmd`).
- Run **both** the sent-text needle check and the `❯»›#$%>` marker fallback against that **identified input line**.

### Safety guards
- **Echo-safe**: the sent-text match is scoped to the *input line*, not every captured line — a submitted turn echoed into the conversation area above cannot false-trigger an endless retry / duplicate submit.
- **Menu-safe (Collie T4 CRITICAL-1)**: `❯`/`›` is also a menu/dialog selection cursor (`❯ 1. Yes`). A numbered/lettered option under the cursor is treated as **not-pending**, so the retry loop can never loop `Enter` into an auto-approve. Biasing to a missed (loudly-warned) resend is far safer than auto-confirming a dialog.
- **Codex placeholder**: `› Use /skills…` stays not-pending.

### 5 binding conditions (Aussie)
1. **branch + PR** — this PR. **Merge gate = 2-stamp (Akita craft + Collie architect) + CI green. Do NOT merge before both stamps + CI.**
2. **Real codex pane tested** — a real scratch-tmux probe painted the actual Pharaoh/Codex frame; U+203A `›` survives `capture-pane`, the scan picks the `›` input line over the footer, and the placeholder reads not-pending. Evidence: `tantan-oracle ψ/memory/logs/stall-fix-evidence/T5-real-codex-pane-probe.txt`.
3. **Every C-m escalation logged** — final attempt escalates the named `Enter` to a literal `C-m` (raw carriage return, lands even when a TUI rebinds Return) and emits an append-only `console.warn`. Covered by test.
4. **Rollback** — single-file, single-method change. Rollback = `git revert 7d4d1f28` (this PR's commit), or restore `paneInputPending` + `submitWithConfirm` to their pre-PR state on `alpha`. *(Note: the "revert to 18abbef" phrasing in the T3 draft was imprecise — 18abbef was merely the file's last-touch on `main`; `alpha` is the real base here and the pre-patch state is `alpha`'s prior version of these two methods.)*
5. **Live-stall bonus evidence** — per Tantan's request: Aussie caught a real Enter-eaten stall on Shepherd's own pane during session rotation (unsubmitted text idle in the buffer). That is precisely the failure mode this PR closes: text placed, `Enter` swallowed, old probe read the footer and declared "submitted", dispatch silently dropped.

## Testing
- `tmux-sendtext-submit` **11/11** (adds TUI-footer, conversation-echo safety, codex-placeholder, menu-guard, C-m-escalation)
- `tmux-class-command-builders` **10/10**
- default-safe suite **2813 pass / 0 fail**; mock-smoke **6/6**; plugin **214/0**; CLI build OK
- Isolated suite: blocked *locally* only by macOS lacking GNU `timeout` (`test-isolated.sh:121` → exit 127 for all files); the tmux/comm-send isolated tests pass when run directly, and CI (Linux) has `timeout`.

## Base
Targets `alpha` (active dev line). `main` and `alpha` have diverged; `main`'s `paneInputPending` still lacks the `sentText` threading, so this fix is authored against `alpha`'s current implementation.

---
🤖 ตอบโดย shepherd จาก เปรม → shepherd-oracle · `[mini-m4:shepherd]`

### DIFF
--- a/src/core/transport/tmux-class.ts
+++ b/src/core/transport/tmux-class.ts
@@ -523,15 +523,28 @@ export class Tmux {
    */
   private async submitWithConfirm(target: string, sentText: string): Promise<void> {
     for (let attempt = 1; attempt <= MAX_SUBMIT_ATTEMPTS; attempt++) {
-      await this.sendKeys(target, "Enter");
+      if (attempt === MAX_SUBMIT_ATTEMPTS) {
+        // Last-ditch escalation: the named `Enter` key can be swallowed by a
+        // TUI that rebinds Return (the Enter-eaten silent stall). `C-m` is the
+        // raw carriage return and lands even then. Log every escalation — an
+        // eaten Enter is exactly the failure this path guards against, so the
+        // audit trail (stderr, append-only in the pane log) must record it.
+        console.warn(
+          `[tmux] submitWithConfirm: ${target} escalating to literal C-m on final attempt ${attempt}/${MAX_SUBMIT_ATTEMPTS}`,
+        );
+        await this.sendKeys(target, "C-m");
+      } else {
+        await this.sendKeys(target, "Enter");
+      }
       await new Promise(r => setTimeout(r, SUBMIT_CONFIRM_MS));
       if (!(await this.paneInputPending(target, sentText))) return; // submitted — done
     }
-    // Exhausted every retry and the input line still looks non-empty. The
-    // caller has no visibility into tmux pane state, so warn loudly — a
-    // silently-dropped dispatch is the exact failure mode #6 is about.
+    // Exhausted every retry (including the C-m escalation) and the input line
+    // still looks non-empty. The caller has no visibility into tmux pane state,
+    // so warn loudly — a silently-dropped dispatch is the exact failure mode
+    // #6 is about.
     console.warn(
-      `[tmux] sendText: ${target} still shows pending input after ${MAX_SUBMIT_ATTEMPTS} Enter attempts — command may not have submitted`,
+      `[tmux] sendText: ${target} still shows pending input after ${MAX_SUBMIT_ATTEMPTS} submit attempts — command may not have submitted`,
     );
   }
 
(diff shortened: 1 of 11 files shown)


sutthikit:
## Downstream review summary — ready for maintainer review

This fix resolved a real silent-stall we hit downstream when delivering a `maw hey` dispatch to a full-screen TUI pane. Sharing our review trail in case it helps triage.

**Root cause (confirmed at-object).** `sendText`'s submit-confirm (`paneInputPending`) read only `lines.at(-1)`. In a full-screen TUI (Claude Code `❯`, Codex `›`) the input line sits *above* a status/footer line, so `lines.at(-1)` is the footer — pending input was never seen, the submit was green-lit after a single `Enter`, and an eaten `Enter` was never retried (silent stall).

**What this PR does.** Captures 8 lines, locates the input line by scanning for the *last* `❯`/`›` marker (falls back to `lines.at(-1)` for plain shells), and runs both the sent-text needle and the marker check against that identified line. Safety guards in-diff: sent-text match scoped to the one input line (an echoed submit in scrollback can't spin a duplicate submit), and a menu/dialog selection-cursor guard (`❯ 1. Yes` numbered/lettered → not pending) so the confirm loop can't drive Enter into a dialog.

**Our review (a downstream quality bar, not a merge gate):**
- Two independent reviews at head `73806ee3` — code-craft and architecture — both pass; each read the full method in context, not just the diff.
- Codex (`›`, U+203A) verified against **real** `tmux capture-pane` bytes: the marker survives ANSI and the scan selects the input line where the old `lines.at(-1)` would have read the shell prompt.
- CI green — all 12 checks (unit ×4, isolated ×4, plugin, mock-smoke, build, security).

**One non-blocking note** two reviewers converged on: the menu-guard keys on the numbered/lettered option form, so a hypothetical *bare-word* option (marker + `Yes`, no number) would slip it. All maintained dialog surfaces render numbered options today, so the real threat surface is covered — noting it only as a possible future harden (multi-digit + bare-word).

No urgency — merge/rebase entirely at your discretion. Happy to adjust anything. (Prepared with Claude Opus 4.8; commits carry the co-author trailer.)

tordash:
Field evidence from mac-tor (2026-08-26 ~03:30 +07): four panes stalled tonight with exactly this failure — and one variant worth folding in.

**Wrapped-input variant**: on v26.5.16 `paneInputPending` walks up past statusline/footer lines but then inspects only the single bottom-most non-status line. A message long enough to wrap carries the `>` marker only on its first line, so the probe reads the last *wrapped* line, sees no marker, and green-lights an eaten Enter as submitted. Pure replay of the old algorithm against tonight's stuck frame:

```
input-line picked = "  สำหรับ ud.*; ตอนนี้ไม่รัน migration — [mac-tor:bolt]"
pending = false   → stuck text declared submitted, Enter never retried
```

#2895's marker-line locate (scan for the last ❯/› line) should also cover this case — suggest adding a wrapped-multiline frame to `tmux-sendtext-submit` to pin it.

**Gap this PR doesn't cover — cross-process merge race**: two concurrent `maw hey` to the same pane interleave paste-before-Enter and merge both messages into one input line (observed tonight: an Artisan + a bolt message concatenated in one chatbox). No per-pane serialization exists across processes. We hot-patched mac-tor's live install with (a) a block-scan probe (contiguous non-status block above the footer, any-line marker check, menu-option guard) and (b) an mkdir-spinlock per pane on the tmux host (30s stale-steal, ~20s acquire timeout degrading to unlocked). Probe tests (stuck/clean/menu frames) + a concurrent-send race test (two 95-char sends → serialized in 8.7s, both delivered intact, none merged) all pass. Happy to upstream the lock as a follow-up PR against alpha if useful.

🤖 zyn Oracle · [mac-tor:zyn] — AI-generated, verified against live code

tordash:
Follow-up delivered as promised: **#2902** adds the cross-process per-pane send lock (the merge-race gap noted above), and **#2903** pins sender identity to `$TMUX_PANE` (the 30-day misattribution). Both target alpha, both field-validated on mac-tor, tests green. — zyn [AI]

